PR #54, opened 2019-02-19
Golden section layout

spadino:
Golden section layout, with only one windows as primary, a second pane and the other stacked as a small rectangle (perfect for terminals during development)

### DIFF
--- a/src/workspacer.Shared/Layout/GoldenLayoutEngine.cs
+++ b/src/workspacer.Shared/Layout/GoldenLayoutEngine.cs
@@ -1,0 +1,97 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace workspacer
+{
+    public class GoldenLayoutEngine : ILayoutEngine
+    {
+        private readonly int _numInPrimary;
+        private readonly double _primaryPercent;
+        
+        private int _numInPrimaryOffset = 0;
+        private double _primaryPercentOffset = 0;
+        
+        public GoldenLayoutEngine() : this(1, 0.618) { }
+
+        public GoldenLayoutEngine(int numInPrimary, double primaryPercent)
+        {
+            _numInPrimary = numInPrimary;
+            _primaryPercent = primaryPercent;            
+        }
+        
+        public string Name => "Golden";
+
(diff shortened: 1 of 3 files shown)


rickbutton:
Hello! I appreciate the contribution!  Apologies for not replying to this sooner!

I am in the middle of finishing my very last class ever at the moment, so I don't have any time to look at this, but I will be done at the end of March! The PR looks good at first glance, and I can test it out and merge it in after that :)

Thanks!!

src/workspacer/ConfigContext.cs
```diff
@@ -45,6 +45,7 @@ public ConfigContext()
 
             _defaultLayouts = () => new ILayoutEngine[] {
                 new TallLayoutEngine(),
+                new GoldenLayoutEngine(),
```
rickbutton: this layout engine looks great! but I don't think that it should necessarily be in the "default" set of layouts.

would you be cool with removing it from the default set? I think that this would make more sense as an optional engine. I'm hoping that I get spend some more time on documentation in the future, which would help expose / show off these engines (that aren't defaults).

src/workspacer.Shared/Layout/GoldenLayoutEngine.cs
```diff
@@ -0,0 +1,97 @@
...
+        
+        public GoldenLayoutEngine() : this(1, 0.618) { }
+
+        public GoldenLayoutEngine(int numInPrimary, double primaryPercent)
+        {
+            _numInPrimary = numInPrimary;
+            _primaryPercent = primaryPercent;            
+        }
+        
+        public string Name => "Golden";
```
rickbutton: this is a minor nitpick, but can this be `golden` instead of `Golden`? so that it matches the other engines

spadino: Hi! Apologize to didn't reply to your suggestion! Of course, I'll work on that and improve it next weeks. 
Many thanks for your work!

src/workspacer/ConfigContext.cs
spadino: Of course!

spadino:
For sure, I'll implement all your suggestions and improvements. Kudos!

spadino:
Closing by now. I'll rework the idea following your requests. ;)

PR #56, opened 2019-03-22
Add a tiling engine for Vertical monitors

ianelletson:
First of all, thanks for this awesome project! Seriously.

Your primary window gets `percent + offset` space, and each additional primary window gets the `percent + offset` of the remaining space. If there are as many primary windows as maximum rows defined, the secondary windows will be auto-minimized.

Secondary windows are tiled in rows across the bottom until the maximum number of rows is met, then each row is split up as needed to fit more windows. Right now, this starts at the topmost secondary row, splitting the row closest to the primary windows.

I think I will add options to instead tile secondary windows from the bottom up and to define how the primary windows' space is determined (possibly through a lambda argument).

This algorithm could almost certainly be substantially improved, so feel free to suggest anything. I may defend a decision but I'm not going to have hard feelings if you don't like something.
I'm also still grokking the whole architecture.

I can add images if this is the type of thing you're interested in merging.


src/workspacer.Bar/Widgets/TitleWidget.cs
```diff
@@ -39,7 +39,7 @@ private IWindow GetWindow()
             var currentWorkspace = Context.WorkspaceContainer.GetWorkspaceForMonitor(Context.Monitor);
             return currentWorkspace.FocusedWindow ??
                    currentWorkspace.LastFocusedWindow ??
-                   currentWorkspace.Windows.FirstOrDefault(w => w.CanLayout);
+                   currentWorkspace.Windows.FirstOrDefault(w => w?.CanLayout ?? false);
```
ianelletson: This fixed an issue where when debugging (with Visual Studio) I would open the workspacer debug window and trigger an exception.

src/workspacer.Shared/Layout/VerticalTiledEngine.cs
```diff
@@ -0,0 +1,152 @@
...
+        private double _primaryPercentOffset;
+
+		public string Name => "vertical";
+
+		public VerticalTiledEngine(int numInPrimary, double primaryPercent, double primaryPercentIncrement, int maxRows)
+		{
+			_numInPrimary = numInPrimary;
+			_primaryPercent = primaryPercent;
+			_primaryPercentIncrement = primaryPercentIncrement;
+            _maxRows = maxRows;
```
ianelletson: Apologies for the spacing. I have a couple different development environments that don't have synced settings. I'll fix this up.

src/workspacer/ConfigContext.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using System.Linq;
 using System.Windows.Forms;
+using workspacer.Shared.Layout;
```
ianelletson: I noticed you used workspacer as the primary namespace. My environment was configured to use namespaces based on folder hierarchy. I can revert this to workspacer or move the remaining layouts to `Shared.Layout`

ianelletson:
Closing this PR until I fix up a bug with primary tiles not utilizing all the available space

rickbutton:
hi! feel free to re-open this when you've got time to fix whatever bugs you find! :)

ianelletson:
Cool! I'll reintegrate the changes you've made into my fork and do that. Haven't fully fixed the bug yet since I rarely encounter it.

PR #57, opened 2019-04-15
Migrated to .NET Core 3

rickbutton:
This works 100%, was mostly a drop in replacement. Waiting on .NET Core 3 to leave preview so that I can officially support this.


josteink:
Just FYI: .NET Core 3 is now officially out. It's actually finally, really out. After all these years :smile:

rickbutton:
Awesome! I'll see if I can snag enough time this weekend to do some final tests and merge it in! Thanks for the poke @josteink !

rickbutton:
no promises, of course, because I'm bad at keeping this project up to date, but the plan is to actually merge this in this weekend, and setup GitHub Actions at the same time (it looks like GitHub actions will be easier to use with dotnet core).

rickbutton:
> but the plan is to actually merge this in this weekend

The weekend ended up being about 8 months long, but I'm finally merging this. I'm very far behind on production releases of this. The *new* plan is to merge this in right now! And then start to address some of the new open issues before doing a new prod release and catching up.

Apologies for the dead repo people! I feel like I've neglected this project. I don't use a windows PC often at home anymore, so I haven't had a need to many any changes here.

PR #73, opened 2019-10-04
Prevent inlining of Logger.Create()

josteink:
Logger.Create() depends on analyzing the call-stack to determine who it is creating a Logger for.

In the -current- workspacer-codebase this probably works fine.

But in my experience, especially in optimized release-builds, some method-calls tend to get inlined, making stack-frame analyzis unreliable, and worst case a cause of bugs or misleading logs.

We can prevent this from ever happening by instructing the compiler to never inline this method-call.

So let's do just that :)


josteink:
FYI: I searched for other places in the codebase with similar use of stack-analysis and this was the only one I found.

PR #74, opened 2019-10-04
Improvements to Color-class

josteink:
- Make it only possible to represent actual colours.
- Make hash-algorithm without chance of collisions.
- Unit-test based on latest Xunit


josteink:
I was *honestly* planning to add some unit-tests for this... In fact I spent most of the time preparing this PR on that bit.

But making a "classic" .NET Framework xUnit-project on Linux/mono and making it run is just too much effort. Really. An amazing amount of non-cool stuff. So I decided to just give up on that for now.

As a contrast, the .NET Core 3.0 stack is amazing, even on Linux and I literally had a xUnit-project up and running in lass than a mnute. Once you've merged the .NET Core 3.0 PR, I'll give adding tests another try :smile:

josteink:
*Now* there should be unit-tests in the solution.

I've also added a unit-test for hash-collisions for the simplest case I could come up with which produced hash-collisions in the original `Color`-class.

One question I have about this PR: For the time being, I've maintained compatibility with the old `Color(int, int, int)` constructor, but that leaves us the possibility of postponing what could be a compile-time error into a runtime error instead.

Is that a really good thing? Or should I just remove the compatibility-constructor so that we know at compile-time that all `Color`-instances are valid?

PR #77, opened 2019-10-08
Make workspacer use reliable, future-proof MSbuild-tasks

josteink:
Using msbuild-tasks may not have OOB support in Visual Studio, but they are way more reliable than regular batch scripts.

With .NET Core 3.0 obsoleting the "regular" pre- and post-build events, you will have to rely on custom targets sooner or later anyway. Might as well make it now :)

An added bonus with this change is that it makes workspacer completely buildable using mono on Linux. The build is now *portable*!

That means you can now setup a free Linux-based CI-build for workspacer on Travis, CircleCI, Azure Pipelines or whatever service you desire, and get free build-validation on all future PRs right here on Github.

And that's pretty darn cool.

Yeah. You're welcome!


josteink:
*Knock knock*

Is this thing still on? :smiley:

rickbutton:
apologies for the delays on these PRs, life gets in the way of open source more than I wish it would!

josteink:
Np. We’re getting there slowly but surely 🙂

PR #79, opened 2019-10-11
Reduce nesting

josteink:
Deeply nested code is hard to read and even harder to reason about.

This PR aims to make the workspacer code more readable by eliminating needless nesting. This is done primarily by inverting existing `if (object != null)`-checks and making exit conditions explicit.

### DIFF
--- a/src/workspacer.Native/Native/WindowsManager.cs
+++ b/src/workspacer.Native/Native/WindowsManager.cs
@@ -95,33 +95,33 @@ namespace workspacer
         {
             var location = Control.MousePosition;
             var handle = Win32.WindowFromPoint(location);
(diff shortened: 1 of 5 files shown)


josteink:
This PR probably has merge conflicts with [this PR](https://github.com/rickbutton/workspacer/pull/69).

From my point of view, it will probably be easier to fix conflicts, if you handle this PR first.

josteink:
This PR probably has merge conflicts with [this PR](https://github.com/rickbutton/workspacer/pull/69).

Right now, this PR is mostly aesthetic, and doesn't *fix* any concrete errors, while the other does. As such, the other PR should clearly have priority before this one.

PR #84, opened 2019-12-02
Avoid causing NPEs during workspacer startup.

josteink:
This closes https://github.com/rickbutton/workspacer/issues/83.


josteink:
Great 🙂

PR #86, opened 2019-12-03
Implement IKeybindManager.Unsubscribe(KeyModifiers mod, Keys key)

josteink:
This addition allows unbinding default-bound keys without resetting the full configuration, as per the documentation.

This closes https://github.com/rickbutton/workspacer/issues/65.


josteink:
Whee!

PR #104, opened 2020-07-04
Add option to configure bar font

JCallicoat:
This PR adds `FontName` option to `BarPluginConfig` with a default of "Consolas".


josteink:
Just looking at the code (haven't tested), I have a few questions:

- Will this break compatibility with existing configs?
- Shouldn't a default font-name be provided by default?

josteink:
Disregard previous comment. LGTM 🙂

rickbutton:
@JCallicoat thanks for the contribution! please feel free to make as many as you want! I don't spend as much time on windows machines lately, so this repo has become dormant, I would love additional contributors!

Issue #111, opened 2020-07-28
Github Code Snippets out of date


N1x0:
All the snippets provided on github are very much out of date and don't work. These should either be updated or completely removed as the updated ones are  available on the website.
If you'd still like to have them on Github I'll make a PR for with the current implementation

josteink:
Hey there and thanks for reporting in.

I think having snippets there is definitely helpful, but not if out of date.

If you're willing to update them, I'm willing to merge :+1:

N1x0:
Alright working on it now, will submit a PR over the weekend once I've figured out some of the quirks.

PR #114, opened 2020-07-29
fix snippet sticky_workspaces.cs

o0morgan0o:
I was having some issues for the configuration of the sticky mode.
I am not familiar with C#, but by looking at the new config guide on the website I managed to make the correction to make it work for me. I assume that the snippet was maybe obsolete. Let me know if it is Ok, i also fixed a typo on the line 19. I think these modifications in this snippet will make it easier for the new users.
Thank you for this great project.

### DIFF
--- a/snippets/sticky_workspaces.cs
+++ b/snippets/sticky_workspaces.cs
@@ -1,22 +1,24 @@
 //g the StickyWorkspaceContainer is useful if you want to emulate workspace modes in awesome/dwm
 //g where workspaces are specifically assigned to monitors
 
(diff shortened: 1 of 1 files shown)


N1x0:
Hi there, 

yes the snippets are quite out of date, I even made an issue about that #111 . I am going to update them all anyways so I could include your fix in that PR if that works for you? Your changes work on my machine as well so seems to be all good.

o0morgan0o:
Yes, it's OK for me. Thanks !

rickbutton (approved):

Issue #92, opened 2020-03-16
Battery Indicator in the bar


MacAhmed:
Hiya,

I have been using workspacer for a while now. I like it quite a bit and have it set up so it basically fills all requirements. The only issue that I have right now is that since I hide the taskbar, I am unable to see the battery indicator. 

Is there a way to show battery percentage in the bar? 

Thankyou.

rickbutton:
this sounds like a great case for a workspacer plugin. currently the docs aren't in a good place for how to do this, but I would gladly accept a PR for this feature.

N1x0:
Hi @MacAhmed , 
Totally agree on your suggestion it'd be a great feature. Now I threw something together which works for me, but it is 
a) not compatible with every device
b) not feature-complete (it just shows a number and not even formatted as percentage)
c) requires additional assemblies (`system.management`)
Once I have worked this out a bit more I'll do a full PR, but for now I hope the code below is somewhat helpful to you.

```csharp

using System;
using System.Collections.Generic;
using System.Text;
using System.Linq;
using System.Management;
using System.Threading.Tasks;
using System.Timers;

namespace workspacer.Bar.Widgets
{
    public class BatteryWidget : BarWidgetBase
    {
        private Timer _timer;
        private int _interval;

        public BatteryWidget(int interval)
        {
            _interval = interval;
        }

        public static string GetBatteryPercentage()
        {
            ObjectQuery query = new ObjectQuery("Select * FROM Win32_Battery");
            ManagementObjectSearcher searcher = new ManagementObjectSearcher(query);

            ManagementObjectCollection collection = searcher.Get();

            foreach (ManagementObject mo in collection)
            {
                foreach (PropertyData property in mo.Properties)
                { 

                    if (property.Name == "EstimatedChargeRemaining")
                    {
                        return property.Value.ToString();
                    };
                }
            }
            return "NaN";
        }
        public BatteryWidget() : this(1000) { }

        public override IBarWidgetPart[] GetParts()
        {
            var percent = GetBatteryPercentage();
            return Parts(percent);
        }

        public override void Initialize()
        {
            _timer = new Timer(_interval);
            _timer.Elapsed += (s, e) => Context.MarkDirty();
            _timer.Enabled = true;
        }
    }
}

```

MacAhmed:
@N1x0 

[...]

PR #116, opened 2020-08-01
Battery Widget

N1x0:
So this is my attempt at solving issue  #92  by creating a battery widget. It uses the `System.Windows.Forms` namespace so it doesn't add any other assemblies to the project. Right now it exposes the following features.

`LowChargeThreshold` - Sets a minimum percentage of the battery charge after which `LowChargeColor` gets applied 
`LowChargeColor` - formats text color when the battery percentage falls below `LowChargeThreshold`
`bool HasBatteryWarning` => enables/disables the `LowChargeThreshold` and associated `LowChargeColor`
`Interval` - refresh rate of the widget

I'd like to update this in the future to also allow the user to display `BatteryChargeStatus`.

### DIFF
--- a/src/workspacer.Bar/Widgets/BatteryWidget.cs
+++ b/src/workspacer.Bar/Widgets/BatteryWidget.cs
@@ -20,22 +20,22 @@ namespace workspacer.Bar.Widgets
         public override IBarWidgetPart[] GetParts()
         {
             PowerStatus pwr = SystemInformation.PowerStatus;
-            float CurrentBatteryCharge = pwr.BatteryLifePercent;
+            float currentBatteryCharge = pwr.BatteryLifePercent;
 
             if (HasBatteryWarning)
             {
-                if (CurrentBatteryCharge <= LowChargeThreshold)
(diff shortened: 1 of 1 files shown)


src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -0,0 +1,50 @@
...
+        public bool HasBatteryWarning { get; set; } = true;
+        public double LowChargeThreshold { get; set; } = 0.10;
+        public int Interval { get; set; } = 5000;
+
+        private System.Timers.Timer _timer;
+
+        public override IBarWidgetPart[] GetParts()
+        {
+            PowerStatus pwr = SystemInformation.PowerStatus;
+            float CurrentBatteryCharge = pwr.BatteryLifePercent;
```
josteink, line 23: Function-local variables should be pascalCased, not CamelCased.

josteink:
LGTM!

@rickbutton : Any objection to this getting merged?

rickbutton:
thank you for this PR @N1x0! Looks great.

thanks @josteink for the review as well :)

PR #118, opened 2020-08-02
Added features to the `TitleWidget` Plugin

N1x0:
Made 2 changes to the `TitleWidget`:

1. The user can now choose to show **only the program name** via the boolean choice of `IsShortTitle`. There are some programs where this does not work as the separator between window contents and programs is not the same everywhere (i.e. MS Teams using ` |` and not `-`), but this seems quite rare.

2. Allows the user to set a custom message for workspaces without any windows, defaults to **"No Windows"**.

### DIFF
--- a/src/workspacer.Bar/Widgets/TitleWidget.cs
+++ b/src/workspacer.Bar/Widgets/TitleWidget.cs
@@ -9,6 +9,8 @@ namespace workspacer.Bar.Widgets
     public class TitleWidget : BarWidgetBase
     {
         public Color MonitorHasFocusColor { get; set; } = Color.Yellow;
+        public bool IsShortTitle { get; set; } = false;
+        public string NoWindowMessage { get; set; } = "No Window";
 
         public override IBarWidgetPart[] GetParts()
         {
@@ -19,10 +21,18 @@ namespace workspacer.Bar.Widgets
 
             if (window != null)
             {
-                return Parts(Part(window.Title, color));
+                if (!IsShortTitle)
+                {
+                    return Parts(Part(window.Title, color));
+                }
+                else 
+                {
+                        string ShortTitle = window.Title.Split("-").Last();
+                        return Parts(Part(ShortTitle, color));
+                }
             } else
             {
-                return Parts(Part("no windows", color));
+                return Parts(Part(NoWindowMessage, color));
             }
         }
 
--- a/src/workspacer.Bar/Widgets/TitleWidget.cs
+++ b/src/workspacer.Bar/Widgets/TitleWidget.cs
@@ -10,7 +10,7 @@ namespace workspacer.Bar.Widgets
     {
         public Color MonitorHasFocusColor { get; set; } = Color.Yellow;
         public bool IsShortTitle { get; set; } = false;
-        public string NoWindowMessage { get; set; } = "No Window";
+        public string NoWindowMessage { get; set; } = "No Windows";
 
         public override IBarWidgetPart[] GetParts()
         {


src/workspacer.Bar/Widgets/TitleWidget.cs
```diff
@@ -9,6 +9,8 @@ namespace workspacer.Bar.Widgets
     public class TitleWidget : BarWidgetBase
     {
         public Color MonitorHasFocusColor { get; set; } = Color.Yellow;
+        public bool IsShortTitle { get; set; } = false;
+        public string NoWindowMessage { get; set; } = "No Window";
```
josteink, line 13: This text used to be "No window**s**" (plural).

N1x0: fixed in last commit, hadn't noticed

josteink:
LGTM.

@rickbutton Any objections to this getting merged?

josteink:
No objections in a month, must surely mean that it's about time to get this merged.

Sorry about the delay!

giri-w:
How can I configure isShortTitle in the configuration. I am not sure where should I add it in this line:
 LeftWidgets  = () => new IBarWidget[] {new WorkspaceWidget(), new TextWidget(": "), new TitleWidget()},
 
 Thanks

flyxyz123:
> 
> 
> How can I configure isShortTitle in the configuration. I am not sure where should I add it in this line:
> LeftWidgets = () => new IBarWidget[] {new WorkspaceWidget(), new TextWidget(": "), new TitleWidget()},
> 
> Thanks

The following code is my approach. It works for the notepad to open a text file with a super long name. However, it doesn't seem to work for a website with a super long name. I'm not familiar with C# but it seems to work partially ok.
https://github.com/flyxyz123/config_local_win10/blob/35c586417f7bad5ab95d85c1bfc631a557b91415/.workspacer/workspacer.config.csx#L26

flyxyz123:
Update:

After further research, it looks like firefox uses '—' character instead of '-' character in the window title string. I rebuild the program after edit the source code to use '—' instead of '-'. After the rebuild, `isShortTtile` is working on firefox but not notepad and edge.

I was thinking maybe we can consider all three characters: '—', '-', and '|' in the source code. Using regex might be a easy approach. However I'm not familar with either regex or C#.

giri-w:
> > How can I configure isShortTitle in the configuration. I am not sure where should I add it in this line:
> > LeftWidgets = () => new IBarWidget[] {new WorkspaceWidget(), new TextWidget(": "), new TitleWidget()},
> > Thanks
> 
> @giri-w The following code is my approach. It works with Notepad and Edge. However, it does not seem to work with Firefox. I am not familiar with C#, but it seems to work partially ok.
> 
> ```
> LeftWidgets = () => new IBarWidget[] { new WorkspaceWidget(), new TextWidget(": "), new TitleWidget(){IsShortTitle = true} },
> ```
> 
> https://github.com/flyxyz123/config_local_win10/blob/35c586417f7bad5ab95d85c1bfc631a557b91415/.workspacer/workspacer.config.csx#L26

Hi @flyxyz123 , thanks a lot for this. I need this for MS edge, so it is fine.

I hope they can add plug in for putting shortcut in the main Bar.

josteink:
> Update:
> 
> After further research, it looks like firefox uses '—' character instead of '-' character in the window title string. I rebuild the program after edit the source code to use '—' instead of '-'. After the rebuild, `isShortTtile` is working on firefox but not notepad and edge.
> 
> I was thinking maybe we can consider all three characters: '—', '-', and '|' in the source code. Using regex might be a easy approach. However I'm not familar with either regex or C#.

That's a good catch. I've created a PR to address that.

https://github.com/rickbutton/workspacer/pull/201

Issue #121, opened 2020-08-12
Duplicate keybinds for: toggle enable/disable workspacer


N1x0:
I was wondering why the `toggle enable/disable` function (which is rather essential IMO) didn't work for me and found that there's a duplicate key binding in `\workspacer\src\workspacer\Keybinds\KeyBindManager.cs` 

 `mod + LShift + E`  is being being assigned to two actions
**Line 209** -  `_context.Enabled = !_context.Enabled, "toggle enable/disable");`
**Line 309** -  `_context.Workspaces.MoveFocusedWindowToMonitor(1), "move focused window to monitor 2");`

The docs on the website also claim that `mod + LShift + E` is used for toggling so might want to edit those as well

josteink:
Good find!￼

Maybe we should also add some warnings to the logs when re-binding already bound keys to easier catch situations like these?

N1x0:
Yeah it definitely should show up somewhere. I'll have a look at how those are done and create a PR to include if @rickbutton agrees

josteink:
No need to wait for Rick. I approve, and I can get PRs merged :smile:

N1x0:
So I looked through the `KeyBindManager` and based on my (very limited) understanding of C# it would require a bit of a rewrite. At the moment key-binds are stored in  the **Dictionary** `_kbdSubs` so there is no real way to check for duplicates as the dictionary does not allow them in the first place.  A possible workaround could be to use a list with something like this:

```csharp
List<KeyValuePair<Sub, NamedBind<KeybindHandler>>> _kbdSubs = new List<KeyValuePair<Sub, NamedBind<KeybindHandler>>>();
```
Then it'd be easy to check for duplicates and add those to the logs as as Debug or even Fatal.

I'm happy to give this a try but not sure if there isn't an easier way or if I've missed something so any comments would be appreciated

rickbutton:
> Yeah it definitely should show up somewhere. I'll have a look at how those are done and create a PR to include if @rickbutton agrees

sounds great, please do!


> So I looked through the `KeyBindManager` and based on my (very limited) understanding of C# it would require a bit of a rewrite. At the moment key-binds are stored in the **Dictionary** `_kbdSubs` so there is no real way to check for duplicates as the dictionary does not allow them in the first place. A possible workaround could be to use a list with something like this:

what if we made `KeyBindManager` throw if you attempt to bind a key that is already bound. this would mean that in order to  rebind a key, you would need to unbind it first. thoughts?

PR #123, opened 2020-08-16
KeybindManager changes 

N1x0:
Quick-fix for #121 
1. added error checking for duplicate assignment, throws exception when a user tries to assign an already used key combination
2. removed unused logger
3. edited default keybind for ` toggle enable/disable` to not conflict with `move focused window to monitor 2`. Should be an intuitive shortcut

### DIFF
--- a/src/workspacer/Keybinds/KeybindManager.cs
+++ b/src/workspacer/Keybinds/KeybindManager.cs
@@ -25,7 +25,6 @@ namespace workspacer
             }
         }
 
-        private Logger Logger = Logger.Create();
 
         private Win32.HookProc _kbdHook;
         private Win32.HookProc _mouseHook;
@@ -59,6 +58,13 @@ namespace workspacer
         public void Subscribe(KeyModifiers mod, Keys key, KeybindHandler handler, string name)
         {
             var sub = new Sub(mod, key);
+            var test = new LauncherResponse();
+            if (_kbdSubs.ContainsKey(sub))
+            {
+                Exception keyCombinationAlreadyAssigned = new Exception($" The Key Combination `{mod}-{key}` is already bound to action: `{name}`");
+
+                _context.QuitWithException(keyCombinationAlreadyAssigned);
+            }
             _kbdSubs[sub] = new NamedBind<KeybindHandler>(handler, name);
         }
 
@@ -205,7 +211,7 @@ namespace workspacer
             Subscribe(MouseEvent.LButtonDown,
                 () => _context.Workspaces.SwitchFocusedMonitorToMouseLocation());
 
-            Subscribe(mod | KeyModifiers.LShift, Keys.E,
+            Subscribe(mod, Keys.Escape,
                 () => _context.Enabled = !_context.Enabled, "toggle enable/disable");
 
             Subscribe(mod | KeyModifiers.LShift, Keys.C,
--- a/src/workspacer/Keybinds/KeybindManager.cs
+++ b/src/workspacer/Keybinds/KeybindManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Drawing;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -25,7 +26,7 @@ namespace workspacer
             }
         }
 
-
+        private Logger Logger = Logger.Create();
         private Win32.HookProc _kbdHook;
         private Win32.HookProc _mouseHook;
 
@@ -34,6 +35,7 @@ namespace workspacer
         private IDictionary<MouseEvent, NamedBind<MouseHandler>> _mouseSubs;
 
         private TextBlockMessage _keybindDialog;
+        private TextBlockMessage _keybindWarning;
 
         public KeybindManager(IConfigContext context)
         {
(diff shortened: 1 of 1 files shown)


josteink:
I think throwing and quitting may be a wee bit radical. 

Some users may want to reassign some actions over existing bindings without setting up all bindings from scratch. 

I think logging s warning, and possibly showing a dialog for this would be better than right out quitting.

N1x0:
I get your view on just quitting being a bit radical, happy to adjust this to  a warning if you think that's better

> Some users may want to reassign some actions over existing bindings without setting up all bindings from scratch. 

This I don't understand as they'd only have to `unsubscribe`  from the duplicate keybind, not all of the defaults.

josteink:
Fair enough, but it will still create increased friction for new users. 

IMO this is somewhat suboptimal for something as "esoteric" and complicated like a tiling window-manager.

src/workspacer/Keybinds/KeybindManager.cs
```diff
@@ -56,9 +57,38 @@ public KeybindManager(IConfigContext context)
             thread.Start();
         }
 
+        public void ShowKeybindWarning(string warningMessage)
```
josteink, line 60: This method doesn't show the dialog per current implementation, it *toggles* it. 

Is that intentional?

src/workspacer/Keybinds/KeybindManager.cs
```diff
@@ -56,9 +57,38 @@ public KeybindManager(IConfigContext context)
...
+        {
+            if (_keybindWarning == null)
+            {
+                _keybindWarning = new TextBlockMessage("workspacer keybinds", "Warning, duplicate keybinds!", warningMessage, new List<Tuple<string, Action>>()
```
josteink, line 64: With this implementation, the warning message provided will only be used the first time and ignored all subsequent times. 

Either pre-allocate this object with a fixed message, or use the provided argument every time when invoked.

src/workspacer/Keybinds/KeybindManager.cs
```diff
@@ -56,9 +57,38 @@ public KeybindManager(IConfigContext context)
...
+            else
+            {
+                _keybindWarning.Show();
+            }
+        }
+
         public void Subscribe(KeyModifiers mod, Keys key, KeybindHandler handler, string name)
         {
             var sub = new Sub(mod, key);
+            var test = new LauncherResponse();
```
josteink, line 83: As far as I can see, this test-variable is never used for anything. 

If so, it should be removed. If not, it needs to be documented with a comment or something why it needs to be here.

N1x0: correct was used for testing, should be removed

src/workspacer/Keybinds/KeybindManager.cs
N1x0: I copied the structure and naming from `ShowKeybindDialog` and retained it for consistency. happy to adjust as it can be simplified to 
```csharp
 public void ShowKeybindWarning(string warningMessage)
        {
            if (_keybindWarning == null)
            {
                _keybindWarning = new TextBlockMessage("workspacer keybinds", "Warning, duplicate keybinds!", warningMessage, new List<Tuple<string, Action>>()
                {
                    new Tuple<string, Action>("ok", () => { }),
                });
          
                _keybindWarning.Show();
            }
        }
```

josteink: That looks better, indeed.

src/workspacer/Keybinds/KeybindManager.cs
```diff
@@ -56,9 +57,28 @@ public KeybindManager(IConfigContext context)
             thread.Start();
         }
 
+        public void ShowKeybindWarning(string warningMessage)
+        {
+            _keybindWarning = new TextBlockMessage("workspacer keybinds", "Warning, duplicate keybinds!", warningMessage, new List<Tuple<string, Action>>()
```
josteink, line 62: We still have the issue with the supplied `warningMessage` only being used for the first invocation here, and then it gets ignored in other subsequent invocations.

Either we should use a fixed message or update `warningMessage` on each invocation.

N1x0: Sorry my programming knowledge is very limited so I don't really understand what you mean by "subsequent invocations" and why the `warningMessage` would get ignored after it was run once. When I bind multiple keys to the same shortcut the `TextBlockMessage` opens up multiple times  with the correct message.

Again sorry for not getting it

josteink: Consider the following C# class:

````c#
public class Buggy
{
    public string LastMessage { get; set; }

    public void SetMessage(string message)
    {
        if (LastMessage == null)
        {
             LastMessage = message;
        }
    }
}
````

Now lets use this class:

````c#
var buggy = new Buggy();
buggy.SetMessage("boo");
buggy.SetMessage("hiss!");

Console.WriteLine(buggy.LastMessage);
// boo
````

Given the name of the method, this is obviously not the expected behaviour of this class, right?

If you look into your code, you have the exact same kind of bug in your implementation of `ShowKeybindWarning()`.

The fix for this problem is either:

- hard-code the message inside the function and make the function don't have any parameters at all, or
- ensure the provided parameter gets used every time, unlike now, where we only use it on the first call.

Do you understand what I mean now?

josteink:
*knock knock*

Is this PR-thing still on? :smile:

rickbutton:
thanks for the ping @josteink . I will review this after work tonight.

Shout out to the small community of people that spend more time working on workspacer than I do!

N1x0:
Hi @josteink  must've been that you hadn't seen my last comment on this. I believe I have already solved the issues you mentioned in your (extremely helpful) message with my latest commit by removing the `   if (LastMessage == null)`  check

josteink:
Indeed. LGTM!

Merged, finally 😀

PR #126, opened 2020-09-16
Colors for different battery levels

haoxiangliew:
Thank you for this awesome project! Makes my time on Windows so much better <3

Credits to Fusion and ianelletson for initial layout code.

Layouts have been tested on configuration and works well.
Battery levels are 10% (red), 50% (yellow), and 100% (green), can be changed if needed.


src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -11,8 +11,12 @@ namespace workspacer.Bar.Widgets
     public class BatteryWidget : BarWidgetBase
     {
         public Color LowChargeColor { get; set; } = Color.Red;
+        public Color MedChargeColor { get; set; } = Color.Yellow;
+        public Color HighChargeColor { get; set; } = Color.Green;
         public bool HasBatteryWarning { get; set; } = true;
         public double LowChargeThreshold { get; set; } = 0.10;
+        public double MedChargeThreshold { get; set; } = 0.50;
+        public double HighChargeThreshold { get; set; } = 1.00;
```
josteink, line 19: Looking at the code below, I would argue `HighChargeThreshold` has no practical value and can be removed.

src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -28,23 +32,26 @@ public override IBarWidgetPart[] GetParts()
...
                     return Parts(Part(currentBatteryCharge.ToString("#0%"), LowChargeColor));
                 }
+                else if (currentBatteryCharge <= MedChargeThreshold)
+                {
+                    return Parts(Part(currentBatteryCharge.ToString("#0%"), MedChargeColor));
+                }
+                else if (currentBatteryCharge <= HighChargeThreshold)
+                {
+                    return Parts(Part(currentBatteryCharge.ToString("#0%"), HighChargeColor));
+                }
```
josteink, line 42: This entire `else if` block can be removed, and `HighChargeColor` should be applied to the final `else` block below.

src/workspacer.Shared/Layout/DwindleLayoutEngine.cs
```diff
@@ -0,0 +1,125 @@
...
+                _numInPrimaryOffset--;
+            }
+        }
+
+        private int GetNumInPrimary()
+        {
+            return _numInPrimary + _numInPrimaryOffset;
+        }
+    }
+}
```
josteink, line 125: This may be a solid new layout for us to consider, but I don't think they belongs in the same PR as one for changing presentation of battery-levels.

Each new feature should IMO (and by Github conventions) have its own PR, so one can independently discuss the new feature and the code implementing it, not to mention merge it on a individual level, per feature.

Please submit layouts as a separate PR, and then that can be discussed there :smiley:

haoxiangliew: I'll open up another PR for layouts

PR #128, opened 2020-09-17
Fix #127: window controls

### LINKED ISSUE #127: Can't close TextBlockMessage
When the keybinds window or an error window pops up there's no way to close it other than using a hotkey (for the keybinds menu) or killing the process from Task Manager.

This can probably be fixed by changing `false` at `workspacer.Share/TextBlockMessage.Designer.cs` line `68` to true.

BeeryShklar:
I could only test the keybinds window so please check the error window or any other windows that may use this class before merge.

### DIFF
--- a/src/workspacer.Bar/Widgets/BatteryWidget.cs
+++ b/src/workspacer.Bar/Widgets/BatteryWidget.cs
@@ -40,13 +40,17 @@ namespace workspacer.Bar.Widgets
                     return Parts(Part(currentBatteryCharge.ToString("#0%"), HighChargeColor));
                 }
             }
-
-            public override void Initialize()
+            else
             {
-                _timer = new System.Timers.Timer(Interval);
-                _timer.Elapsed += (s, e) => Context.MarkDirty();
-                _timer.Enabled = true;
+                return Parts(currentBatteryCharge.ToString("#0%"));
             }
+        }
+
+        public override void Initialize()
+        {
+            _timer = new System.Timers.Timer(Interval);
+            _timer.Elapsed += (s, e) => Context.MarkDirty();
+            _timer.Enabled = true;
         }
     }
 }


josteink:
Fine by me. Any objections from anyone?

src/workspacer.Shared/TextBlockMessage.Designer.cs
```diff
@@ -1,89 +1,89 @@
...
+			this.ButtonLayout.Name = "ButtonLayout";
+			this.ButtonLayout.Size = new System.Drawing.Size(757, 30);
+			this.ButtonLayout.TabIndex = 2;
+			// 
+			// TextBlockMessage
+			// 
+			this.AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
+			this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
+			this.ClientSize = new System.Drawing.Size(784, 461);
+			this.ControlBox = true;
```
BeeryShklar, line 68: This is the line I changed

rickbutton (changes requested):
hi! thanks for the PR, a couple of notes:

- The `TextBlockMessage` window actually has buttons to close the window/take an action, but it seems like they are broken/hidden for some reason. I would rather we fix the layout of this dialog box than enable the control buttons.
- Looks like spaces got converted to tabs in this file, do you mind reverting back to spaces? Might need to tweak your editor config.

BeeryShklar:
Changed the indentation.
Other than that I don't think I can help :(
I did notice that as @N1x0 said, when you double click the title a button that can close the window appears.

N1x0:
Yeah so I've noticed quite a while ago but never mentioned it (my bad!). I think it can be fixed by looking at the `TextBlockMessage` form itself. I'm a bit occupied at the moment but will give it a shot over the weekend.

PR #129, opened 2020-09-17
Colors for different battery levels

haoxiangliew:
Current levels are set to 10% (Red) 50% (Yellow) and 100% (Green)


src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -11,8 +11,12 @@ namespace workspacer.Bar.Widgets
     public class BatteryWidget : BarWidgetBase
     {
         public Color LowChargeColor { get; set; } = Color.Red;
+        public Color MedChargeColor { get; set; } = Color.Yellow;
+        public Color HighChargeColor { get; set; } = Color.Green;
         public bool HasBatteryWarning { get; set; } = true;
         public double LowChargeThreshold { get; set; } = 0.10;
+        public double MedChargeThreshold { get; set; } = 0.50;
+        public double HighChargeThreshold { get; set; } = 1.00;
```
josteink, line 19: This value isn't used for anything and should be removed. 

Apart from that, this all looks good to me 👍

josteink (changes requested):

src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -28,23 +32,26 @@ public override IBarWidgetPart[] GetParts()
                 {
                     return Parts(Part(currentBatteryCharge.ToString("#0%"), LowChargeColor));
                 }
+                else if (currentBatteryCharge <= MedChargeThreshold)
+                {
+                    return Parts(Part(currentBatteryCharge.ToString("#0%"), MedChargeColor));
+                }
+                else if (currentBatteryCharge <= HighChargeThreshold)
```
josteink, line 39: Same comment as last time: we don't need high charge threshold, and this else-is block should be merged with the following else-block.

josteink (changes requested):

src/workspacer.Bar/Widgets/BatteryWidget.cs
haoxiangliew: Changed it to only have a Low Battery Color

josteink:
I think maybe you removed too much? This is the entire PR now: :smile: 

image

haoxiangliew:
Just changed the file, and removed the redundant else function. Already in stable is the low battery color (red), my commit only adds two levels of color to the battery widget for medium battery (less than 50%) and high battery (above 50%)

haoxiangliew:
Also the new layouts are in another PR now :D

josteink:
Looks much better!

Before we merge, could you please remove the "HighChargeThreshold" variable, since it is no longer in use :)

haoxiangliew:
Done!

josteink:
LGTM!

Thanks for the patch!

josteink (approved):

haoxiangliew:
No problem. thanks for reviewing my PR!

rickbutton:
thanks for the PR @haoxiangliew and thanks for reviewing @josteink

josteink:
Np, Rick. 

As for the layouts PR, I'll leave that to someone else. Don't think that's my area of expertise 😉

PR #131, opened 2020-09-18
Bugfix BatteryWidget.cs

haoxiangliew:
The else loop is not redundant since it is an option to the BatteryWidget.Parts() class
Just tested this configuration on my setup, earlier PR had a bug when diff with my local setup :(


josteink:
OK. I merged this and encouraged this change, so this is clearly my bad.

Sorry for doing sloppy code-review :grinning:

Issue #136, opened 2020-10-26
Grid layout


BeeryShklar:
I really want to have a grid layout but I don't even know where to start, can someone help me with that? I saw the layouts in `shared/layout` but I didn't really know how to turn them into a grid layout.
Also would be nice having a grid layout in the default layouts.

Thanks!

N1x0:
Hi @BeeryShklar , I really like the idea of a grid-based layout so I tried to come up with something. Sadly though actually implementing it is a massive headache so what I wrote is not a "true" grid, because instead of dynamically sizing depending on the number of windows (e.g. 9 windows --> 3x3  or 16 windows --> 4x4). it just adds another *column* to the right and has a fixed number of *rows* (2). I am trying to create a true dynamic grid but as I said it's quite the challenge.

Also please note that none of the resizing functions are implemented because IMO that kind of defeats the purpose of grid. So what I'd suggest is to load with the `FullLayoutEngine` as secondary layout. That way you can always maximize the focused window.

`sticky.CreateWorkspace("GRID", new GridLayoutEngine(), new FullLayoutEngine());`

Let me know what you think 😄 

```csharp
 public class GridLayoutEngine : ILayoutEngine
    {
       
        public GridLayoutEngine() {}

        public string Name => "grid";

        public IEnumerable<IWindowLocation> CalcLayout(IEnumerable<IWindow> windows, int spaceWidth, int spaceHeight)
        {
            var list = new List<IWindowLocation>();
            var numWindows = windows.Count();
            int rows;
            int cols;

            if (numWindows == 0)
                return list;

            if (numWindows < 4)
            {
                cols = numWindows;
                rows = 1;
            }
            else
            {
                cols = numWindows % 2 == 0 ? numWindows / 2 : (numWindows + 1) / 2;
                rows = 2;
            }
      

            for (int i = 0; i < numWindows; i++)
            {
            
                if (i < cols)
                {
                    list.Add(new WindowLocation(spaceWidth / cols * i, 0, spaceWidth / cols, spaceHeight / rows, WindowState.Normal));
                }
                else
                {
                    list.Add(new WindowLocation(spaceWidth / cols * (i - cols), spaceHeight / rows, spaceWidth / cols, spaceHeight / rows, WindowState.Normal));
                }
            
            }

            return list;

        }

        public void ShrinkPrimaryArea() { }
        public void ExpandPrimaryArea() { }
        public void ResetPrimaryArea() { }
[...]

PR #141, opened 2020-11-22
Two implementations of Grid-based layout engines

N1x0:
This is to address issue #136 by adding two new `LayoutEngines`. One is a *pseudo-grid* with a fixed number of rows (2),while the other scales dynamically i.e. it adjusts the number of rows as well.

I have opted to keep my first solution (pseudo grid) because it looks nicer for a certain number of windows (5, 7,11) and I couldn't figure out a neat way to structure the "true" Grid the same way.

As stated in the original issue none of the (re)-sizing functions are implemented so what I'd suggest is to load the workspace with the `FullLayoutEngine` as secondary layout
*e.g.*
`sticky.CreateWorkspace("GRID", new GridLayoutEngine(), new FullLayoutEngine());`

Open to any comments and suggestions 😄


josteink:
I'm no  master of custom layouts honestly, but I did skim over the code and it looks good to me.

If I would have any suggestions it would be to use less "clever" and inlined code for calculating the number of rows and columns. They are kinda hard to read as is.

N1x0:
> 
> 
> I'm no master of custom layouts honestly, but I did skim over the code and it looks good to me.
> 
> If I would have any suggestions it would be to use less "clever" and inlined code for calculating the number of rows and columns. They are kinda hard to read as is.

Thanks for your quick reply @josteink , I adjusted the ternary  to be a proper `if-else` block for the pseudo grid and rewrote `Math.Pow(numWindows, 0.5)` to `Math.Sqrt(numWindows)` for the true grid which should make it a bit more readable.

Happy if @BeeryShklar could give them a spin too since he originally made the request for this addition.

src/workspacer.Shared/Layout/GridLayoutEngine.cs
```diff
@@ -0,0 +1,53 @@
...
+      
+
+                int rows = (int)(Math.Sqrt(numWindows) + 0.5);
+
+                while (numWindows % rows != 0)
+                {
+                    rows -= 1;
+                }
+
+                int cols = numWindows / rows;
```
josteink, line 29: These lines seems to be indented 4 characters too much, making them appear as if they are part of the if-block.

N1x0: my bad, fixed in newest commit

josteink:
LGTM with latest changes. :+1:

Any objections from anyone else?

BeeryShklar:
@N1x0  sorry for the late response, I tried running your fork and couldn't.
This is what I have done:
I added your fork as a remote, made a new branch and cloned your `GridLayouts` branch.
When I ran `dotnet run` from `src/workspacer` it failed.
This is the log:
```
D:\code\rickbutton\workspacer\src\workspacer\workspacer.csproj(59,5): error MSB3030: Could not copy the file "D:\code\rickbutton\workspacer\\src\workspacer.Bar\bin\Debug\netcoreapp3.1\win-x64\workspacer.Bar.dll" because it was not found.
```
Can someone help me run it? I am *very* new to dotnet and workspacer is the first time i have actually coded in c#.

N1x0:
Hi @BeeryShklar  thanks for wanting to try it out. So the error message is pretty clear I think, no? Doesn't seem like the `.dll` file was found. I tried the same for me and it worked, but also my `workspacer.bar.dll` was in a separate folder. I'd recommend just using visual studio instead for debugging and testing via `dotnet run`

rickbutton:
Thanks for the contribution! I'd love to have more grid-based layouts available.

I think that we may be able to utilize the increment/decrement methods in order to determine the number of rows in the grid. Does anyone like that idea?

josteink:
That sounds like a good idea to me. I'm all for it.

josteink (approved):
LGTM. We can eventually add the resize => row functionionality later on, if people ask for it.

josteink:
If @rickbutton has approved, there's really no good reason not to merge this.

This has what... Been hanging around for 3-4 months now? :fearful: 

Let's get it done!

josteink:
If rickbutton also likes this, there's really no good reason not to merge this. This has what... Been hanging around for 3-4 months now? :fearful: 

But it seems like at least 2 reviews are required now. Holding on until @rickbutton or @dalyIsaac get's around to approve.

dalyIsaac:
I quite like the pseudo-grid. I do find some of the transitions to columns in the dynamically-scaling grid are slightly disconcerting, which I think @N1x0 was referring to.

The logic probably could be updated in the future, particularly with the resize => row functionality.

However, **it's an awesome start.**

LGTM

dalyIsaac (approved):

josteink:
"At least 2 approving reviews are required by reviewers with write access."

Oh well. Guess it's just me and @rickbutton then?

PR #143, opened 2020-11-25
DwindleLayout

Fusion:
Hi. I saw the pull request by @haoxiangliew  and am happy to see he credited me :)

However, for some reason, his pull request seems to have lost some important bits from the gist he copied it from.

So, here it is, my original gist, but as a built-in class instead.


N1x0:
Hi @Fusion and thanks for the PR,

I've just tested it and it looks good overall. Only comment is that there is a maximum of 5-6 windows (depending on the application) for the secondary zone before stuff gets moved off-screen and things start looking weird. So at that stage you need to increase the number of windows in the primary zone. Not really a concern for me because I rarely have that many windows on a single workspace but just something that may need a note of some sorts.

Only other comment would be the name `dwindle`, just a personal preference but `golden ratio` or `fibonacci` sounds cooler (even though they are probably less descriptive).

Happy to see the number of `LayoutEngines` grow 😄 

**6 Windows in secondary zone**
Annotation 2020-11-25 211621

**7th window gets moved off-screen**
Annotation 2020-11-25 211117

**Added another window to primary zone**
Annotation 2020-11-25 211621

**Added a third window to primary zone**
Annotation 2020-11-25 211756

Fusion:
Thanks for your feedback @N1x0.

Regarding the layout's name, there is a fairly pedestrian consideration: I believe that what is commonly referred to as a Fibonacci layout entails a layout folding on itself, like a shell. This layout is different is that while it creates smaller panels, it follows this pattern: `left-right-bottom-right-bottom-...` as opposed to `left-right-bottom-left-up-right-bottom-left-up-...`

I'll be honest: I have no idea what folks would prefer. I cannot, myself, decide which one I like best :/ so let me know what you would prefer in the end.

Regarding secondary windows, I believe that you have a better understanding of the logic than I do. I tried to reproduce your example, with 2 primary windows, and created  a dozen secondary windows, but none of them were moved off screen. This could be due to using a rather large display? At any rate, regarding writing a note on that topic I'd say that you are right, but mostly because I had enough time myself to forget all about the fantastic "primary area" feature ;)

josteink:
PRs haven’t been merged in *months* and Rick haven’t been answering emails either. 

Right now it seems a fork is needed to keep workspacer going forwards. 

I’m discussing this in a issue here: https://github.com/rickbutton/workspacer/issues/211. Feel free to let me know your thoughts.

N1x0 (approved):
I can write something in the docs about the particular behaviour, the code is otherwise fine.

dalyIsaac (approved):
LGTM!

PR #146, opened 2020-11-25
Fix the gap by calculating offset

MSylvia:
As discussed here: https://github.com/rickbutton/workspacer/issues/139, this PR fixes issues where gaps are present when laying out windows.

I now have it where all the gaps are basically gone using a combination of `GetWindowRect` and `DwmGetWindowAttribute` to compute an offset to apply to the final window position.

With this change windows that previously had gaps on the Left, Right and Bottom (like Explorer, Firefox, Edge, and Terminal) no longer do.

### DIFF
--- a/src/workspacer.Native/Native/WindowsDeferPosHandle.cs
+++ b/src/workspacer.Native/Native/WindowsDeferPosHandle.cs
@@ -67,7 +67,12 @@ namespace workspacer
                 _toNormal.Add(window);
             }
 
-            Win32.DeferWindowPos(_info, window.Handle, IntPtr.Zero, location.X, location.Y, location.Width, location.Height, flags);
+            int X = location.X + window.Offset.X;
+            int Y = location.Y + window.Offset.Y;
+            int Width = location.Width + window.Offset.Width;
+            int Height = location.Height + window.Offset.Height;
+
+            Win32.DeferWindowPos(_info, window.Handle, IntPtr.Zero, X, Y, Width, Height, flags);
         }
     }
 }
--- a/src/workspacer.Native/Native/WindowsDeferPosHandle.cs
+++ b/src/workspacer.Native/Native/WindowsDeferPosHandle.cs
@@ -67,10 +67,12 @@ namespace workspacer
                 _toNormal.Add(window);
             }
 
-            int X = location.X + window.Offset.X;
-            int Y = location.Y + window.Offset.Y;
-            int Width = location.Width + window.Offset.Width;
-            int Height = location.Height + window.Offset.Height;
+            // Calculate final position for window
+            var offset = window.Offset;
+            int X = location.X + offset.X;
+            int Y = location.Y + offset.Y;
+            int Width = location.Width + offset.Width;
+            int Height = location.Height + offset.Height;
 
             Win32.DeferWindowPos(_info, window.Handle, IntPtr.Zero, X, Y, Width, Height, flags);
         }
--- a/src/workspacer.Native/Native/WindowsWindow.cs
+++ b/src/workspacer.Native/Native/WindowsWindow.cs
@@ -78,7 +78,8 @@ namespace workspacer
             {
                 var buffer = new StringBuilder(255);
                 Win32.GetClassName(_handle, buffer, buffer.Capacity + 1);
-                return buffer.ToString();            }
+                return buffer.ToString();
+            }
         }
 
         public IWindowLocation Location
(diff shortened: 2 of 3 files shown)


N1x0 (commented):
Tested it and works well (mind the 1px gap)

src/workspacer.Native/Native/WindowsWindow.cs
```diff
@@ -98,7 +99,55 @@ public IWindowLocation Location
...
+                int Width2 = rect2.Right - rect2.Left;
+                int Height2 = rect2.Bottom - rect2.Top;
+
+                // Offset
+                int X = X1 - X2;
+                int Y = Y1 - Y2;
+                int Width = Width1 - Width2;
+                int Height = Height1 - Height2;
+
+                return new WindowLocation(X, Y, Width, Height, state);
```
josteink, line 150: I realize it's a matter of re-using an existing class for a new purpose, but it seems a bit strange to go fetching a `Offset` property in your code and being given a `WindowLocation`.

Could a [`System.Drawing.Rectangle`](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.rectangle?view=net-5.0) be used instead for some of these cases? Would that be better?

I'm honestly not sure myself, but some discussion around this may still be useful :smiley:

src/workspacer.Native/Native/WindowsDeferPosHandle.cs
```diff
@@ -67,7 +67,12 @@ public void DeferWindowPos(IWindow window, IWindowLocation location)
                 _toNormal.Add(window);
             }
 
-            Win32.DeferWindowPos(_info, window.Handle, IntPtr.Zero, location.X, location.Y, location.Width, location.Height, flags);
+            int X = location.X + window.Offset.X;
+            int Y = location.Y + window.Offset.Y;
+            int Width = location.Width + window.Offset.Width;
+            int Height = location.Height + window.Offset.Height;
```
josteink, line 73: I guess this is the "meat" of the fix which removes the gaps. But reading the code, it's not entirely obvious *how*.

My intuitive understanding would be that to remove the gaps we need to make the values of `X` and `Y` smaller (and to do that we should *subtract* something), while we want to make `Width` and `Height` bigger (by *adding* something).

Is my understanding correct?

If so, here we are adding in both cases, and there is no subtraction to the `X` and `Y` variables. Is `Offset.X` and `Offset.Y` expected to return a negative values, while `Offset.Width` and `Offset.Height` return positive ones? 

This section could benefit from values/forms better matching the mental model of how one fills the gaps, or by adding more comments/documentation outlining what is supposed to happen here.

josteink: Also: In this section you are pulling in `Offset` 4 times to access 4 sub-properties. This means you actually calculate the full Offset variable once for each and every one of those sub-properties.

Assign `Offset` to a local variable and access sub-properties from that one instance instead.

josteink (changes requested):
Overall it looks good, but I do have a few comments on style, naming and actual implementation.

Let me know what you think :)

src/workspacer.Native/Native/WindowsWindow.cs
```diff
@@ -98,7 +99,55 @@ public IWindowLocation Location
                     state = WindowState.Maximized;
                 }
 
-                return new WindowLocation(rect.Left, rect.Top, rect.Right - rect.Left, rect.Bottom - rect.Top, state);
+                int X = rect.Left;
+                int Y = rect.Top;
+                int Width = rect.Right - rect.Left;
+                int Height = rect.Bottom - rect.Top;
+
+                return new WindowLocation(X, Y, Width, Height, state);
```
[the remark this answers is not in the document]
josteink: No big deal. It was mostly a question of if there might bea bug hiding  in here or not.

src/workspacer.Native/Native/WindowsWindow.cs
```diff
@@ -98,7 +99,55 @@ public IWindowLocation Location
...
+            {
+                WindowState state = WindowState.Normal;
+                if (IsMinimized)
+                {
+                    state = WindowState.Minimized;
+                }
+                else if (IsMaximized)
+                {
+                    state = WindowState.Maximized;
+                }
```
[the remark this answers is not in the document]
josteink: Great!

src/workspacer.Native/Native/WindowsWindow.cs
josteink: Much better now. I also like how we don't need to handle `state` anymore. :)

src/workspacer.Native/Native/WindowsDeferPosHandle.cs
josteink: This is clearly much better now. I still think it's a little unintuitive to *add* the X and Y offsets, but on the other hand, it wouldn't make much sense either to return a rectangle where we change up the X and Y without ... Also adjusting the height and width.

So yeah. Let's keep it as is then.

josteink (approved):

rickbutton (approved):
awesome work!

rickbutton:
thanks @josteink for the review

PR #149, opened 2020-12-03
System Tray Icon is removed properly on exit

OldKros:
I noticed when restarting Workspacer or quitting through the right click function on the System Tray icon, that the icon wasn't properly being disposed as you can see by this image.
Screenshot 2020-12-03 111033

Looking through the code, this function was implemented in SystemTrayManager but just wasn't being called on quit. 
Additionally, if Workspacer is killed via task manager or pressing stop whilst debugging, the Icon remains. However testing this with other tray utilising programs (keepassxc and VLC I tested) the same thing happened with them.

### DIFF
--- a/src/workspacer/ConfigContext.cs
+++ b/src/workspacer/ConfigContext.cs
@@ -148,7 +148,7 @@ namespace workspacer
             var str = JsonConvert.SerializeObject(response);
             _pipeServer.SendResponse(str);
         }
-        
+
         public void Restart()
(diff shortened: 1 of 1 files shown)


josteink:
LGTM!

josteink (approved):

josteink:
Thanks for the patch. Consider it merged!

rickbutton:
nice catch

PR #154, opened 2020-12-10
Override update branch with settings.yaml

### LINKED ISSUE #153: Automatic Update issues
So workspacer should have automatic updates which is provided through the `AutoUpdater.NET` `NuGet` package. However I'm not sure if that is working as intended. 

The following is implemented in `\workspacer\src\workspacer\Program.cs`:

```csharp
#if BRANCH_unstable
            var branch = "unstable";
#elif BRANCH_stable 
            var branch = "stable";
#else
            string branch = null;
#endif
            if (branch != null)
            {
                var xmlUrl = $"https://workspacer.blob.core.windows.net/installers/{branch}.xml";

                AutoUpdater.ApplicationExitEvent += QuitForUpdate;

                System.Timers.Timer timer = new System.Timers.Timer(1000 * 60 * 60);
                timer.Elapsed += (s, e) =>
                {
                    AutoUpdater.Start(xmlUrl);
                };
                timer.Enabled = true;
                AutoUpdater.Start(xmlUrl);
            }
```

Maybe I didn't look carefully enough but I couldn't find a preprocessor directive for either `BRANCH_unstable` or `BRANCH_stable` meaning the check for updates never happens because `branch`will always evaluate to `null`. I tested this by editing the Assembly Version and the update does trigger correctly when the directives are present. I still don't quite understand the purpose of the 1hr timer though...

A second potential issue is that the user has no way of configuring this behaviour without going into the source files. I think it'd be nice to have this accessible through the config file instead. 

A a side note would implementing checksums for the downloads be possible? Would be nice if there's a way verify the integrity of the download prior to execution.

Happy to write something up for this if there's interest to have this feature.

sitiom:
Resolves #153.

You can now simply override or switch from the defined branch with your own by setting `~/.workspacer/settings.yaml` with:
```yaml
branch: stable
# or
branch: unstable
```

Edit:
Added `none`  as an option to disable auto-update. This would be useful for installing from [scoop](https://github.com/lukesampson/scoop-extras/pull/5278).

### DIFF
--- a/src/workspacer.Shared/Branch.cs
+++ b/src/workspacer.Shared/Branch.cs
@@ -1,0 +1,18 @@
+﻿namespace workspacer
+{
+    public enum Branch
+    {
+        /// <summary>
+        /// No branch (disable updates)
+        /// </summary>
+        None,
+        /// <summary>
+        /// The stable branch
+        /// </summary>
+        Stable,
+        /// <summary>
+        /// The unstable branch
+        /// </summary>
+        Unstable
+    }
+}
--- a/src/workspacer.Shared/IConfigContext.cs
+++ b/src/workspacer.Shared/IConfigContext.cs
@@ -48,6 +48,11 @@ namespace workspacer
         IEnumerable<ILayoutEngine> ProxyLayouts(IEnumerable<ILayoutEngine> layouts);
 
         /// <summary>
+        /// The branch to check updates with
+        /// </summary>
+        Branch? Branch { get; set; }
+
+        /// <summary>
         /// ConsoleLogLevel controls the log level of the debug output window (alt-shift-i by default)
         /// </summary>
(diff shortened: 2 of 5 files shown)


rickbutton:
We already have an optional config file where you can write custom C# to change settings. Would it make more sense to just allow you to write a line of code in that file that overrides the update branch?

sitiom:
> We already have an optional config file where you can write custom C# to change settings. Would it make more sense to just allow you to write a line of code in that file that overrides the update branch?

Makes sense.

sitiom:
Now you could do:

```cs
context.Branch = Branch.Stable;
// or
context.Branch = Branch.Unstable;
// or disable updates
context.Branch = Branch.None;
```

sitiom:
Closing this in favor of #156.

Issue #152, opened 2020-12-10
Support for IconFonts (Bar)


N1x0:
So I've played around a bit with trying to get icons into the Bar and have had some degree of success. I've used [this](https://github.com/dkrusky/FontAwesome-For-WinForms-CSharp) to go about it. Basically all there was to it was to add the `Fonts.cs` file to `workspacer.Bar` solution and add the `.ttf` filepath. Edit: the Font needs to be installed system-wide as well. Just double clicking will call the installer so it's not a lot of additional work. Potentially the font could be packaged as content/resource file to circumvent this.

From there you can add icons to any part of the bar by just providing the name `Fonts.fa.Icon_Name`. This works for workspaces too!
![Annotation 2020-12-10 140712](https://user-images.githubusercontent.com/47623143/101733891-1bedd480-3af2-11eb-92e1-8bd7532e7933.jpg)

Now this only works if that font is selected for the entire bar (e.g. setting ` FontName = "FontAwesome"`), so that's a bit of drag, but I personally like having icons even if they're just eye-candy so I'd be happy to play with it a bit more and see if it's possible to have two fonts.

Curious to hear what people think, if it's interesting enough I'd be happy to make a PR

josteink:
Font-awesome licensing seems a bit complex: https://github.com/FortAwesome/Font-Awesome/blob/master/LICENSE.txt

Is it OK to bundle it with workspacer as is?

N1x0:
I hadn't really considered that, from a first glance there don't seem to be any issues but I think I'll have to read it thoroughly to be sure.

lauverityhill:
> Curious to hear what people think, if it's interesting enough I'd be happy to make a PR
I'm interested, please do!

lauverityhill:
> Curious to hear what people think, if it's interesting enough I'd be happy to make a PR

I'm interested, please do!

N1x0:
After some more testing there's two things necessary to make this work well I think. First the font file currently used is quite old so I'll prepare an equivalent with the much more recent Font Awesome 5 Free version. 
Secondly, and more important, to do this have a proper implementation it's necessary that each `Widget` can be assigned it's own font. Otherwise all widgets will take the icon font and the result looks weird (screenshot below)

![image](https://user-images.githubusercontent.com/47623143/101872033-2cb44e00-3bb7-11eb-988e-cfe575658399.png)

[...]

PR #155, opened 2020-12-11
Individual widget fonts

N1x0:
First step to enable icon fonts as mentioned in #152 . This PR allows each widget to set its own font and effectively overwrite the default font of the bar.

### DIFF
--- a/src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
+++ b/src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
@@ -12,14 +12,12 @@ namespace workspacer.Bar.Widgets
       
         private Timer _timer;
 
-        public ActiveLayoutWidget()
-        {
-        }
+        public ActiveLayoutWidget() { }
 
         public override IBarWidgetPart[] GetParts()
         {
             var currentWorkspace = Context.WorkspaceContainer.GetWorkspaceForMonitor(Context.Monitor);
-            return Parts(Part("[" + currentWorkspace.LayoutName + "]", fore: null, back: null, () =>
+            return Parts(Part("[" + currentWorkspace.LayoutName + "]", partClicked: () =>
             {
                 Context.Workspaces.FocusedWorkspace.NextLayoutEngine();
             }, fontname: FontName));
--- a/src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
+++ b/src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
@@ -22,7 +22,7 @@ namespace workspacer.Bar.Widgets
             return Parts(Part("[" + currentWorkspace.LayoutName + "]", fore: null, back: null, () =>
             {
                 Context.Workspaces.FocusedWorkspace.NextLayoutEngine();
-            }, FontName));
+            }, fontname: FontName));
         }
 
         public override void Initialize()
--- a/src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
+++ b/src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
@@ -9,6 +9,7 @@ namespace workspacer.Bar.Widgets
 {
     public class ActiveLayoutWidget : BarWidgetBase
     {
+        public string FontName { get; set; } = null;
         private Timer _timer;
 
         public ActiveLayoutWidget()
@@ -21,7 +22,7 @@ namespace workspacer.Bar.Widgets
             return Parts(Part("[" + currentWorkspace.LayoutName + "]", null, null, () =>
             {
                 Context.Workspaces.FocusedWorkspace.NextLayoutEngine();
-            }));
+            }, FontName));
         }
 
         public override void Initialize()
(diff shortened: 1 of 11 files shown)


src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
```diff
@@ -9,6 +9,7 @@ namespace workspacer.Bar.Widgets
 {
     public class ActiveLayoutWidget : BarWidgetBase
     {
+        public string FontName { get; set; } = null;
```
josteink, line 12: Any reason not to just put this in BarWidgetBase, if all widgets are going to have this?

src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -17,6 +17,7 @@ public class BatteryWidget : BarWidgetBase
         public double LowChargeThreshold { get; set; } = 0.10;
         public double MedChargeThreshold { get; set; } = 0.50;
         public int Interval { get; set; } = 5000;
+        public string FontName { get; set; } = null;
```
josteink, line 20: Same about this declaration. Why not just inherit it from base instead?

src/workspacer.Bar/Widgets/BatteryWidget.cs
```diff
@@ -29,20 +30,20 @@ public override IBarWidgetPart[] GetParts()
             {
                 if (currentBatteryCharge <= LowChargeThreshold)
                 {
-                    return Parts(Part(currentBatteryCharge.ToString("#0%"), LowChargeColor));
+                    return Parts(Part(currentBatteryCharge.ToString("#0%"), LowChargeColor, null, null, FontName));
```
josteink, line 33: These invocation might look cleaner if using [named arguments](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments).

src/workspacer.Bar/Widgets/FocusedMonitorWidget.cs
```diff
@@ -8,14 +8,16 @@ namespace workspacer.Bar.Widgets
 {
     public class FocusedMonitorWidget : BarWidgetBase
     {
+        public string FontName { get; set; } = null;
```
josteink, line 11: This could be inherited too.

src/workspacer.Bar/Widgets/TextWidget.cs
```diff
@@ -8,6 +8,7 @@ namespace workspacer.Bar.Widgets
 {
     public class TextWidget : BarWidgetBase
     {
+        public string FontName { get; set; } = null;
```
josteink, line 11: And this one.

src/workspacer.Bar/Widgets/TextWidget.cs
```diff
@@ -17,7 +18,7 @@ public TextWidget(string text)
 
         public override IBarWidgetPart[] GetParts()
         {
-            return Parts(_text);
+            return Parts(Part(_text,null,null,null,FontName));
```
josteink, line 21: This invocation would *definitely* look nicer with named arguments.

src/workspacer.Bar/Widgets/TimeWidget.cs
```diff
@@ -9,6 +9,7 @@ namespace workspacer.Bar.Widgets
 {
     public class TimeWidget : BarWidgetBase
     {
+        public string FontName { get; set; } = null;
```
josteink, line 12: Inherit this one from BarWidgetBase too.

josteink:
I've taken a quick look and it mostly looks OK to me, but there's a bit code-repetition which can be removed, and there's some function-invocations which does look a bit risky, which IMO can also be improved.

Fix that and I'll take a second look :smiley:

src/workspacer.Bar/Widgets/ActiveLayoutWidget.cs
N1x0: Blanked on the fact that this was possible... Thanks for pointing it out!

N1x0:
Updated based on your feedback 😊

src/workspacer.Bar/Widgets/TextWidget.cs
```diff
@@ -18,7 +18,7 @@ public TextWidget(string text)
 
         public override IBarWidgetPart[] GetParts()
         {
-            return Parts(Part(_text,null,null,null,FontName));
+            return Parts(Part(_text, fore: null, back: null, partClicked: null, FontName));
```
josteink, line 21: IN this case I was more thinking eliminating redundant parameters using named arguments. Like this:

````c#
return Parts(Part(_text, fontName: FontName));
````

File: src/workspacer.Bar/Widgets/WorkspaceWidget.cs [hunk 9]
```diff
@@ -14,7 +14,7 @@ public class WorkspaceWidget : BarWidgetBase
         public Color WorkspaceEmptyColor { get; set; } = Color.Gray;
         public Color WorkspaceIndicatingBackColor { get; set; } = Color.Teal;
         public int BlinkPeriod { get; set; } = 1000;
-        public string FontName { get; set; } = null;
+        //public string FontName { get; set; }
```
josteink, line 17: This line should be removed, not commented out.

src/workspacer.Bar/Widgets/TextWidget.cs
N1x0: oh my bad, totally misunderstood you there

josteink:
That looks much better. Thanks!￼

Consider it merged!￼

josteink (approved):

PR #130, opened 2020-09-17
New Layouts (Golden Ratio & Vertical List)

haoxiangliew:
Credits to Fusion and ianelletson for initial code

These configurations have been tested on my setup, and I haven't encountered any bugs

PR #161, opened 2020-12-22
Customise Gaps per layout 

OldKros:
Hi again, slight bigger PR this time, I have added the ability to customise the gap between windows. 
I have added default key bindings as mod-shift-h to decrease and mod-shift-l to increase the gaps.
This feature is added per layout engine and would need to be implemented in each one - Tall layout is the one I use mostly so I have done that one although the Full layout will be next to do so lookout for a PR if this one is suitable. Default gap size is 10 pixels although I feel it should be 0 and then people can either specify a gap size in their config or manually increase it everytime to their liking.
This PR will break any custom LayoutEngines anyone has added or the PRs (#143 and #130) as it adds two functions to ILayoutEngine which need to be implemented in each LayoutEngine. 
I have set the increment to 2 pixels at a time, as I initially thought to half the gap for each window that are next to each other so the full gap would still be in effect which would be hard to do with an odd number if the increment was 1.
Where you see `gapSize * 2` for the width and height this is needed because 1 gapSize would counter the effect of the gap offset from the opposite side of the window, so we need to remove another gapSize amount of pixels for it to look even.

Initially I wanted to add it to WindowsWindow.cs where the actual window size is calculated although it doesn't know whether the window is primary, secondary, first, or last, which makes deciding if we want to have a full gap (between a window and the edge of the screen) or half gap or zero gap difficult to impossible. (still not sure which method is best to use between windows (full right gap on primary and 0 left gap on secondary is what I have opted for)).

This is my second (First really, the other was a one-liner) PR ever and any feedback would be welcome.
Im still learning the intricacies of Git and Github so if anything is off do let me know.
Additionally, I know rick doesn't have much time to work on this but I'm sure along with others I'm feeling this project slowly becoming great and with this change I think I can finally switch from bug.n.


rickbutton:
Hi! Thanks for this PR! I will love this feature.

Your implementation will work, but may I suggest a possible alternative implementation? `IConfigContext::AddLayoutProxy` allows you to wrap layout engines in another layout engine, which is how the gap for the menu bar is implemented (see `MenuBarLayoutEngine`). You can look at `workspacer.bar.ConfigContextExtensions::AddBar` for an example of how this is used.

My theory is that you can implement configurable gaps as a layout proxy instead. That would prevent the need for each layout engine to handle gaps themselves, and would mean that we could put this feature in a new plugin `workspacer.Gaps` (if we wanted, not necessarily sold on the current project structure).

OldKros:
Hi Rick,
Firstly, I agree that this implementation is a quick and dirty method and not the most effective solution since it would need to be implemented in existing, as well as new layout engines. I imagine your proposed solution would be effective at creating the gaps on the outside of the workspace (see red areas). However, the gaps around the windows (green areas) would be something that, at the moment, I am not entirely sure how to do properly.
Some methods I have are:
1. tell the layout engine to shrink the window size by a certain amount (kinda like my implementation here although less messy and less edge cases - it would shrink every window on every side instead of certain edges for certain windows depending where they are in the layout) 
1. have some kind of GapManager in `workspacer.Gaps` that resizes the windows to create gaps after they have been laid out by the LayoutEngine (with no gaps).

Workspacer_Gap_ideas


In any case I will sit on this until I hear back or if anyone comes up with a superior solution.

josteink:
Hey @OldKros.

There seems to be demand for what this PR delivers, so I'm trying to give it some attention :laughing: 

That said, it seems there's some merge conflicts with latest changes in git master.

Can you fix those so that the CI build will run, and we can take it from there? :smile:

OldKros:
Hi @josteink, I actually don't have time to work on this at the moment, looks like the concept has been picked up by someone else #193 that looks to be what I would turn this into anyway.
Feel free to close this PR if #193 is merged.

josteink:
It looks like #193 will serve our needs fine. Thanks for putting in the effort, none the less!

PR #163, opened 2020-12-22
Added the ability to cycle between monitors

OldKros:
I found the swapping between monitors with dedicated keybinds for each monitor rather annoying and instead wanted one keybind to go to the next monitor (or another to go to the previous) however that ability was not implemented. It was for the workspaces but not for monitors which I thought was a bit weird, so here we are.


N1x0 (changes requested):
Hi @OldKros, nice PR! 

I'm doing some testing at the moment and it seems to work well and quite frankly I feel like this could not only extend the current way of handling monitor interaction but actually replace in part, at least for focusing monitors. For moving windows to other monitors I believe it makes sense to retain the original approach as well. Just my two cents,  it depends on what @rickbutton  thinks in the end.

What's missing is adding these actions to the KeybindManager as default subscriptions. E.g.
```csharp
            Subscribe(mod | KeyModifiers.LShift, Keys.Right
                () => _context.Workspaces.SwitchFocusToNextMonitor(), "Focus next monitor");

            Subscribe(mod | KeyModifiers.LShift, Keys.Left
                () => _context.Workspaces.SwitchFocusToPreviousMonitor(), "Focus previous monitor");

            Subscribe(mod | KeyModifiers.Control, Keys.Right
                () => _context.Workspaces.MoveFocusedWindowToNextMonitor(), "move focused window to next monitor");

            Subscribe(mod | KeyModifiers.Control, Keys.Left
                () => _context.Workspaces.MoveFocusedWindowToPreviousMonitor(), "move focused window to previous monitor");
```

Potentially some more minor cleanups but other than that a great addition IMO!

N1x0 (approved):

josteink (approved):

josteink:
Yeah this looks good to me to.

Sorry about taking so long time getting this reviewed. Consider it merged!

PR #170, opened 2021-01-06
Fixed Bar sizing issue

NeoNlAmp3:
Issue mentioned in [https://github.com/rickbutton/workspacer/issues/169](url)

### DIFF
--- a/src/workspacer.Bar/BarForm.cs
+++ b/src/workspacer.Bar/BarForm.cs
@@ -65,9 +65,9 @@ namespace workspacer.Bar
         private void OnLoad(object sender, EventArgs e)
         {
             this.Height = _config.BarHeight;
-            this.Width = _monitor.Width;
             var titleBarHeight = this.ClientRectangle.Height - this.Height;
             this.Location = new Point(_monitor.X, _monitor.Y - titleBarHeight);
+            this.Width = _monitor.Width;
             _timer.Enabled = true;
         }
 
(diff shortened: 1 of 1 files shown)


josteink:
I really can't see why this change should fix your issues. And I kinda don't like fixing bugs by making changes which in theory shouldn't ... change things. Basically I don't like magic :smile: 

On the other hand I can't see how this change can cause any other negative adverse effects, so unless anyone else objects, I'm OK with merging this.

NeoNlAmp3:
Yes I was also quite confused.
But this morning I realized that there is a gap in Y direction.
So I checked what is happening again.

The Gap
Scaling
Scaling2
Monitor Settup

It hit me when I saw 24 to 30 => 30/24 = 1,25 which is the Windows scaling of Monitor 1 and 3. Monitor 2 is the only one at a scaling on 100%.
Basically this.Location = new Point(....) will cause some scaling after the width and height are set and change them to x/125

I don't know how this could be prevented as Point is a class inside System.Windows.Forms.Control and I can't debug into it to see who is the baddy

NeoNlAmp3:
I updated the Branch.

Now the size will be overwritten after the scaling happened. It's not a good solution an quite ugly in my opinions but it works so far. 
Maybe some one else can test the scaling behavior to confirm the issue/fix.

josteink:
I notice you set the `Height` only, and then later set both `Height` and `Width`.

Is that intentional or a slip?

NeoNlAmp3:
It's intentional, if the Height is not set before the Location is assigned with Point, the bar will slide up by the difference of the scaling. So in my case the bar was 6px up outside of the screen.
So the height must be set before the scaling happens, then the scaling does it stuff and then I reset the height to the intended value.

josteink (approved):

josteink:
Oh well. Consider it merged then :smile:

PR #177, opened 2021-02-04
Floating layout engine

lvlts:
Hi,

I have implemented a Floating Layout Engine for those who would like to use such a layout. I always keep a workspace with a floating layout and use it for applications which do not behave as expected in other layout types.

I hope you also find it useful.

**Note:** the floating layout reads the `Window` properties (`Location`, `Offset`) in their original form. Therefore, if a Bar (or more) are defined, it needs to know the height of the Bar(s). By default, this is `30`, and this is what the normal constructor does. If you have a different Bar height (e.g. `50`) you want to call it in your configuration like so:

```
context.DefaultLayouts = () => new ILayoutEngine[] { new FullLayoutEngine(), new TallLayoutEngine(), new FloatingLayoutEngine(50) }
```


josteink:
People in general seem kinda hesitant to review new layouts... I guess that means they don't feel the need for too many?

That said, this PR might help out people who are having [issues with typical JetBrains products](https://github.com/rickbutton/workspacer/issues/61) like Rider or CLion?

Maybe you can pull some of those users in here and see if it helps their case?

As for the code itself:

> Note: the floating layout reads the Window properties (Location, Offset) in their original form. Therefore, if a Bar (or more) are defined, it needs to know the height of the Bar(s). By default, this is 30, and this is what the normal constructor does. If you have a different Bar height (e.g. 50) you want to call it in your configuration like so

Don't we have a [`BarPluginConfig`](https://github.com/rickbutton/workspacer/blob/master/src/workspacer.Bar/BarPluginConfig.cs)-object type which could be used here instead, somehow?

src/workspacer.Shared/Layout/FloatingLayoutEngine.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var windowList = windows.ToList();
+
+            for (var i = 0; i < numWindows; i++)
+            {
+                var window = windowList[i];
+
+                list.Add(new WindowLocation(
+                            window.Location.X-window.Offset.X,
+                            window.Location.Y-window.Offset.Y-_barHeight,
```
dalyIsaac, line 34: I've encountered an issue with a secondary monitor to the right of the primary monitor, where a window will continually be shifted further to the right and thus hidden.

I think it's because the `WindowLocation`'s `X` and `Y` positions are being updated here:
https://github.com/rickbutton/workspacer/blob/0aaff4437222e471ea0ffd282b948a9dfb8b7ff5/src/workspacer.Shared/Workspace/Workspace.cs#L353-L354

I think this can be fixed by:
```suggestion
                            0, 0,
```

That also means `_barHeight` isn't necessary.

lvlts: I currently don't have multiple monitors, so it will take me a few days until I can test it.

The exact issue you've experienced happens out of the box if I don't use the `_barHeight` variable using a single monitor setup. That single variable was the reason why the floating layout worked in the single monitor setup.

Your suggestion will snap the window to the top left corner of the screen(?). I'll take a deeper look at the multi-monitor floating setup to make sure it works with more monitors.

Rather, to prevent shifting I think the right "equation" would be:

`window.Location.Y-window.Offset.Y- (<number of monitors> * _barHeight)`

rickbutton:
Thanks for the PR!

A suggestion for this change: instead of returning a list of window locations that are just "the same location that each window is currently at", I wonder if it is a cleaner solution to instead provide a way for a layout engine to tell workspacer that a window should just have it's location ignored entirely, and only care about keeping it inside the current workspace. Thoughts?

josteink:
This PR has been stale since @rickbutton's review/comment earlier on.

Does anyone of the current maintainers have an opinion on if this is OK now in it's current form, or if we should close this PR due to inactivity?

dalyIsaac:
Closed due to inactivity

Issue #61, opened 2019-05-23
Rider (Jetbrains C# IDE) windows flash on any non-editor interaction


ianelletson:
Upon opening any pop up or clicking any menu item, the Rider window flashes. Sadly, it is incredibly annoying and in all honesty I think Workspacer might be more important to me than Rider.

MacAhmed:
This is an issue that I have recently also experienced with IntelliJ, where the code-editor getting focus results in a flash of the window. The bug is present on both the laptop screen, and the external monitors. 

Has anyone found a workaround by any chance?

ianelletson:
My only workaround was `Alt+T` (or your shortcut for un-managing a window), unfortunately. I made a workspace that I kept empty aside from the IDE.

I haven't used Rider since I started using VS 2019 so I don't know if and how things have changed.

rickbutton:
>This is an issue that I have recently also experienced with IntelliJ, where the code-editor getting focus results in a flash of the window

Can you describe the flashing in more detail? I've seen something that might be similar to this with WinSCP.

oxysoft:
Same issue here for Rider, window flashes almost with every character typed

stianeklund:
This is definitely an issue in the latest version as well.
It also seems to not be exclusive to Jetbrains as I can see "flickering" that are contained to lists in other windows while the Jetbrains IDE is open.

Please see the attached gif demonstrating the issue
![S1m565oWqO](https://user-images.githubusercontent.com/4148195/81389616-6bae8c80-911a-11ea-9656-1137f85a4499.gif)

mdkozlowski:
do you see any way to eventually fix the flickering? part of the reason why I installed workspacer was after getting a bigger screen, to be able to tile my IDE with other windows on the same workspace

Blucknote:
https://user-images.githubusercontent.com/15163043/103132484-c7338780-46b5-11eb-8098-add42a0bd001.mp4

May be it would be helpful, IntelliJ also flicks when mouse cursor transit between windows

PR #178, opened 2021-02-10
Remove title change event to prevent flickering in some apps

lvlts:
This PR removes the (so far) unused TITLE CHANGE event, which was causing flickering effects on several apps (editors):

- Notepad
- JetBrains products (in particular, the text editing window)
- Some Sun AWT based products

**Note:** this is not a complete fix for JetBrains issues (#61) or Sun AWT. However, it makes the IDE and other applications very usable, with the additional filter added in your configuration (in particular, for dropdowns, menus and other floating windows in Sun AWT):

```
    context.WindowRouter.AddFilter((window) => !window.Class.Contains("SunAwtWindow"));
```


josteink:
Good job. Looks good to me :)

josteink (approved):

josteink:
>  with the additional filter added in your configuration (in particular, for dropdowns, menus and other floating windows in Sun AWT)

Would it make sense to add that as a default-filter to make this more usable out of the box?

Maybe you could send a PR for that too? :smiley:

PR #181, opened 2021-02-14
Add more default filters

lvlts:
This PR adds an additional default filter for window classes containing `SunAwtWindow`.

This drastically reduces the flickering for all Sun AWT-based applications, such as JetBrains IDEs, especially when switching context menus or in autocomplete window scenarios. 

Issue #61 should be mostly mitigated, to the point it becomes comfortable to use with workspace, when used along with the previous related pull request (#178).


josteink (approved):
LGTM

josteink:
Thanks for the patch! Consider it merged!

PR #183, opened 2021-02-15
Fixes wrong workspace for window after restarting.

dalyIsaac:
## Reproduction:

Let there be 3 workspaces.
- Open Notepad on the 2nd workspace.
- Switch back to the first workspace.
- Restart workspacer.

## Expected behavior:

Notepad opens on the 2nd workspace.

## Actual behavior:

Notepad opens on the 3rd workspace.

## Diagnosis

This occurs because `StickyWorkspaceContainer._workspaces` is mutated in
`StickyWorkspaceContainer.AssignWorkspaceToMonitor`.

https://github.com/rickbutton/workspacer/blob/0aaff4437222e471ea0ffd282b948a9dfb8b7ff5/src/workspacer.Shared/Workspace/StickyWorkspaceContainer.cs#L112-L113

### DIFF
--- a/src/workspacer.Shared/Workspace/StickyWorkspaceContainer.cs
+++ b/src/workspacer.Shared/Workspace/StickyWorkspaceContainer.cs
@@ -58,7 +58,7 @@ namespace workspacer
         {
             CreateWorkspace(_context.MonitorContainer.FocusedMonitor, name, layouts);
         }
-        
+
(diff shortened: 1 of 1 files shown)


N1x0 (approved):
Tested and works just as intended from what I can tell. Very nice fix to a long-standing bug, 
Thanks a lot!

rickbutton (approved):

PR #186, opened 2021-02-19
Move focused window to adjacent Workspace and switch to that Workspace

dlundgaard:
Hi, first time making a pull request, so please do point out if this PR is not up to standard.

What I'm proposing here is adding functionality to move the focused window to an adjacent workspace and "follow it" in the sense that you switch to the workspace where the focused window now resides. 

As such, it is just a "shorthand" for placing a window in the `n`'th workspace with `Alt+Shift+{n}` and then switching to it with `Alt+{n}`.

I added the following key bindings for the above mentioned, for which I could not find any conflicts and I believe are quite intuitive as an extension of the original ones:
- `Alt+Shift+Left`    Move Focused Window To Previous Workspace And Switch to It
- `Alt+Shift+Right`  Move Focused Window To Next Workspace And Switch to It

### DIFF
--- a/src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
+++ b/src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
@@ -16,6 +16,8 @@ namespace workspacer
 
         IWorkspace GetNextWorkspace(IWorkspace currentWorkspace);
         IWorkspace GetPreviousWorkspace(IWorkspace currentWorkspace);
+        int GetNextWorkspaceIndex(IWorkspace currentWorkspace);
+        int GetPreviousWorkspaceIndex(IWorkspace currentWorkspace);
 
         IWorkspace GetWorkspaceAtIndex(IWorkspace currentWorkspace, int index);
 
--- a/src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
+++ b/src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
@@ -20,6 +20,7 @@ namespace workspacer
         int GetPreviousWorkspaceIndex(IWorkspace currentWorkspace);
 
         IWorkspace GetWorkspaceAtIndex(IWorkspace currentWorkspace, int index);
+        int GetWorkspaceIndex(IWorkspace workspace);
 
         IWorkspace GetWorkspaceForMonitor(IMonitor monitor);
         IMonitor GetCurrentMonitorForWorkspace(IWorkspace workspace);
--- a/src/workspacer.Shared/Workspace/IWorkspaceManager.cs
+++ b/src/workspacer.Shared/Workspace/IWorkspaceManager.cs
@@ -29,6 +29,8 @@ namespace workspacer
         void SwitchFocusToPreviousMonitor();
         void SwitchFocusedMonitorToMouseLocation();
         void MoveFocusedWindowToWorkspace(int index);
+        void MoveFocusedWindowAndSwitchToNextWorkspace();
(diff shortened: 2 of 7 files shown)


josteink:
I haven't looked too closely at the code, but we just merged this PR here today:

https://github.com/rickbutton/workspacer/pull/163

Would you consider that an adequate solution to your needs, or does your PR target other use-cases?

I'm in no way trying to discourage you from contributing, but just trying to be sure we don't merge duplicate functionality.

dlundgaard:
Thanks for your reply, @josteink!

I had a look at #163, and if I'm not mistaken, that one relates to the shuffling/cycling of monitors, whereas this (#186) is meant to streamline moving around the focused window between workspaces on the same monitor. 

However, it seems that the key bindings are overlapping, so some new ones will be needed if they are to be included as default. A suggestion would be `Alt+Ctrl+Left`, `Alt+Ctrl+Right`, but I would love to hear if you have any convention or guidelines on how to pick them so it stays intuitive.

josteink:
To be honest, since I was able to go full Linux on my workstation at work, I'm no longer really using Windows to such an extent that I need/use workspacer anymore. 

I just still hang around and do code-review and stuff because I'm very much sympathetic to the goals of the project 😊

So that means @rickbutton or some other actual regular users will have to voice opinions about the ergonomics and preferences for new key-bindings.

src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
```diff
@@ -16,6 +16,8 @@ public interface IWorkspaceContainer
 
         IWorkspace GetNextWorkspace(IWorkspace currentWorkspace);
         IWorkspace GetPreviousWorkspace(IWorkspace currentWorkspace);
+        int GetNextWorkspaceIndex(IWorkspace currentWorkspace);
```
dalyIsaac, line 19: Since there's no guarantee that this method will be called with the _actual_ current workspace, I personally think that it would make more sense to use a name like `workspace` or `targetWorkspace`.

Similarly for `GetPreviousWorkspaceIndex`.

src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
```diff
@@ -16,6 +16,8 @@ public interface IWorkspaceContainer
...
+        int GetPreviousWorkspaceIndex(IWorkspace currentWorkspace);
```
dalyIsaac, line 21: Since `GetNextWorkspaceIndex` and `GetPreviousWorkspaceIndex` are being added, I think it would make sense to expose a `GetWorkspaceIndex(IWorkspace workspace)` method too.

dalyIsaac (commented):
This functionality seems useful, and the new key-bindings make sense to me.

src/workspacer.Shared/Workspace/IWorkspaceContainer.cs
dlundgaard: I agree. I updated the default key bindings and added `GetWorkspaceIndex(IWorkspace workspace)` in 5de1d6c.

dalyIsaac: Awesome! Would you be able to use `GetWorkspaceIndex` in `GetNextWorkspaceIndex` and `GetPreviousWorkspaceIndex` too?

dalyIsaac (approved):

josteink (approved):

josteink:
LGTM. Thanks for the patch!

Feel free to send more PRs in the future.

PR #188, opened 2021-02-20
ShowKeybindDialog now renders a table.

dalyIsaac:
This adds `KeyValueTable`, which is used to replace `TextBlockMessage` for displaying the current keybindings.

**Changes:**

- Added `KeyValueTable`
- Added `Capitalize` string extension method
- `GetKeybindString` capitalizes keys to aid readability
- `GetKeybindString` adds support for `KeyModifiers.Win`

image

### DIFF
--- a/src/workspacer/Keybinds/KeybindManager.cs
+++ b/src/workspacer/Keybinds/KeybindManager.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
@@ -34,7 +34,7 @@ namespace workspacer
         private IDictionary<Sub, NamedBind<KeybindHandler>> _kbdSubs;
         private IDictionary<MouseEvent, NamedBind<MouseHandler>> _mouseSubs;
 
-        private TextBlockMessage _keybindDialog;
+        private KeyValueTable _keybindTable;
         private TextBlockMessage _keybindWarning;
 
         public KeybindManager(IConfigContext context)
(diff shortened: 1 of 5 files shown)


N1x0:
I like the change and the addition of search functionality is very nice! Tested it and all key binds show and search works for both the description as well as the shortcuts. One issue however, is that if you open the dialog once and then close it you will run into an "object disposed" error if you try to open it again. I'd either include some error handling for this or change the way the KeyValueTable gets disposed. If you can fix this it'd be a fantastic change IMO. Thanks for the PR

dalyIsaac:
Thanks for pointing out that error - it should be fixed now.

src/workspacer/Keybinds/KeybindManager.cs
```diff
@@ -375,45 +375,54 @@ private string GetKeybindString(KeyModifiers mods, Keys keys)
...
+                parts.Add("Alt");
             if (mods.HasFlag(KeyModifiers.LShift))
-                parts.Add("shift");
+                parts.Add("Shift");
 
             if (keys == Keys.Oemcomma) { parts.Add(","); }
             else if (keys == Keys.OemPeriod) { parts.Add("."); }
             else if (keys == Keys.Oem2) { parts.Add("/"); }
-            else if (new Regex("d\\d").IsMatch(keys.ToString().ToLower()))
+            else if (new Regex("d\\d").IsMatch(keys.ToString().Capitalize()))
```
N1x0, line 388: Currently because of the additional "d" in the regex this will print numbered keybinds as "DN" e.g. "D1"
I would change this to:
```csharp
else if(new Regex("\\d").IsMatch(keys.ToString().Capitalize()))
```

By removing the additional "d" it just returns the number which makes it cleaner IMO

N1x0 (changes requested):
The fix for for the KeyValueTable works as intended. Would like to add 1 minor change to the  regex but other than that all looks good!

N1x0 (approved):
All good! Will have to wait for @josteink  or @rickbutton  for the merge though

josteink (approved):
That's awesome. Thanks for the patch!

PR #189, opened 2021-02-21
Ignores Snip & Sketch.

dalyIsaac:
This means that Snip & Sketch can be used as normal, instead of being treated as a separate window.

This can be tested by calling Snip & Sketch using the shortcut `Win` + `Shift` + `S`


N1x0 (approved):
Had this in my config since day one. Definitely a worthwhile addition.

N1x0 (approved):

josteink (approved):

josteink:
Thanks for improving workspacer 🙂

PR #190, opened 2021-02-21
Allows `IWorkspace` instances to be renamed.

dalyIsaac:


josteink:
Looks OK to me. Any possible unintended consequences?

dalyIsaac:
I've had a look, and I don't think there's any side effects.

josteink (approved):

josteink:
Consider it merged then 😁

PR #197, opened 2021-03-18
Layout engines now ignore minimized windows

### LINKED ISSUE #72: Unable to minimize windows
I'm unable to minimize windows with workspacer running. Any idea why?

dalyIsaac:
Resolves #72


josteink:
Given @rickbutton 's reply on the original issue:

> So far this is intentional. Workspacer overrides the "minimize"/"show" actions of the default window manager. i.e. it will send a "show" message to a minimized window as soon as you click minimize.

... I guess this change should probably be enabled by a config-setting and not a hard-coded default like here.

Julynx (approved):

JulioCabria (approved):

dalyIsaac:
I've updated it to so that `IConfigContext` has a Boolean `CanMinimizeWindows`. However, I'm not sure if this is the best place for it in the config.

The default behaviour is now to not allow minimizing.

josteink (approved):
Looks good to me!

josteink:
@rickbutton  What do you think about this PR as it looks now?

josteink:
Can anyone else (with write-access) review and approve this PR? It needs one more approval.

N1x0 (commented):
Code looks good, would only ask to also update the docs & default config file with the additional flag to make sure users are aware of it's existence.

N1x0 (approved):
LGTM!

PR #204, opened 2021-05-09
Added ability to customize text to some of the widgets

schiltz3:
ActiveLayoutWidget - Can now set left and right padding
FocusedMonitorWidget - Can now set focused and unfocused text (Using FontAwsome fonts allows for icons to be used here)
WorkspaceWidget - Can now set left and right padding


schiltz3:
Found Buggs

PR #214, opened 2021-06-27
Referenced template config, better AddBar example

system32uwu:
Referenced template config file containing the #r directives (in my case the template file was never copied to `%USERPROFILE%/.workspacer`, may be a bug), so in case someone deleted the file or it didn't got generated like in my case know where to look for it.

Added a more complete example on how to use the AddBar function, it's not covered that for example Right/LeftWidgets expects a function.


docs/content/quickstart.md
```diff
@@ -39,7 +39,7 @@ C:\Users\<username>\.workspacer\workspacer.config.csx
...
 
-workspacer's config file allows for __type checking__ and __intellisense__! currently, only VSCode with the C# extension is supported. opening the `.workspacer` folder in your home directory inside VSCode with the C# extension installed will provide syntax highlighting, type checking, and autocomplete automatically because of the `#r` directives at the top of the config file (so don't remove those, or else this won't work!).
+workspacer's config file allows for __type checking__ and __intellisense__! currently, only VSCode with the C# extension is supported. opening the `.workspacer` folder in your home directory inside VSCode with the C# extension installed will provide syntax highlighting, type checking, and autocomplete automatically because of the [`#r` directives at the top of the config file](https://github.com/workspacer/workspacer/blob/495ab72fdf80d8aedbdd790cc638e5f8e97a476e/src/workspacer/Config/workspacer.config.template.csx#L1) (so don't remove those, or else this won't work!).
```
dalyIsaac, line 42: Should it be a permanent link?

system32uwu: Would it be better to just reference the entire file then?

dalyIsaac: I think so - the benefit would be that if the template is updated, then the link doesn't need to be updated.

On the other hand, if the template is moved, that'll break the link (hopefully this would be picked up in review) 🤞 .

system32uwu: Ok, I'll change it

dalyIsaac (approved):
LGTM!

system32uwu:
> 
> 
> LGTM!

Nice :), I was thinking about adding a section showcasing users' custom configs like [mine](https://github.com/system32uwu/dotfiles/blob/main/windows10/.workspacer/workspacer.config.csx), in case someone prefers reading usable examples instead of the entire docs, which we'll of course sooner or later expand, but just so people can get started quickly using workspacer.

dalyIsaac:
> I was thinking about adding a section showcasing users' custom configs like mine

Yeah that'd be awesome! My config is at https://github.com/dalyIsaac/workspacer-config/blob/main/workspacer.config.csx. It's a bit bloated - hopefully I can cut down on its size and merge some things into workspacer itself.

system32uwu:
Your config is very useful, it'll be a great reference to link in the docs since it has everything a new user will be interested in. I'll make a new PR including this section I mentioned, if you know someone else that has their configs public we can reference them as well.

N1x0 (approved):
LGTM!

josteink:
> I was thinking about adding a section showcasing users' custom configs like mine

Couldn't we just use a wiki for that?

N1x0:
> > I was thinking about adding a section showcasing users' custom configs like mine
> 
> Couldn't we just use a wiki for that?

@josteink  Could and probably should! Can you enable the wiki for the repo?

josteink:
It's all yours, guys: https://github.com/workspacer/workspacer/wiki

PR #215, opened 2021-06-27
Add configurable padding and focus text for some widgets

schiltz3:
## Add attributes to the standard widgets allowing their text and padding to be modified in the config. 

NOTE: The added attributes have default values corresponding to their original hard-coded values, so this is backward compatible.

---
### 1. Edit or remove padding characters around edges of widget elements:

**Before:** &ensp;&ensp; image

**After:** &ensp;&ensp;&ensp; image

### 2. Change active monitor character and active layout padding

**Before:** &ensp;&ensp; image

**After:** &ensp;&ensp;&ensp; image


N1x0:
I like the ability to set padding, I'd just move it to *IBarWidgetPart* and then implement in the individual widgets as necessary. Also IMO defaults should not have any padding, I know this is up to personal preference, just stating mine.

If you could make the necessary changes I'll approve them, thanks for the PR!

schiltz3:
I believe I made the changes you suggested. 

The default padding has been removed. I can add it back in if we think it is outside the scope of this PR. 

I did not change the Focus/Unfocus text, but I think it could be changed to the focus text being the text attribute, and hard coding unfocused to be an empty string.

LMK what you think

N1x0 (approved):
Changes look good to me, I also agree that FocusedMonitorWidget should return something meaningful instead of just '****' or "", but I think that's better kept in a seperate PR. I will write something in the docs to reflect the change. Please have a look there as well as the author ;)

dalyIsaac (approved):
I agree with @N1x0 - otherwise LGTM!

schiltz3:
> Changes look good to me, I also agree that FocusedMonitorWidget should return something meaningful instead of just '****' or "".

 I may update the docs later with examples of how to use Extended Unicode with workspacer. With fonts like [FontAwsome](https://fontawesome.com/v5.15/icons?d=gallery&p=2&s=brands,regular,solid)

lauverityhill:
I love this PR because since day 1 I've always customized the padding (and needed to recompile workspacer everytime it updates).

However, a padding set to an empty string (`""`) instead of with a space (" ") leads to this behavior:

https://user-images.githubusercontent.com/68158964/123742181-d3570680-d8dd-11eb-927c-4b1493f2e8ce.mp4

because of an inconsistency in line 88 of WorkspaceWidget.cs:
` return visible ? LeftPadding + workspace.Name + RightPadding : $" {workspace.Name} ";`

If both padding variables are `""`, then once you focus on a workspace its length shrinks on the bar.


I would've liked to submit a fix myself, but I'm still not comfortable with Git and programming in general.

PR #218, opened 2021-06-28
Handles empty titles for short titles

dalyIsaac:
**Current behaviour:** `Last` raises `InvalidOperationException` when the `Enumerable` has length of 0.

**Fixed behaviour:** `Last` is only called when `parts` has a length > 0.


josteink:
If this area is causing errors, I think maybe we should try isolating it in a function which we can easily add unit-tests for?

dalyIsaac:
Good idea - I'll do that

N1x0 (approved):
LGTM!

josteink (approved):

PR #221, opened 2021-06-29
Ignores Windows 11 start and search

dalyIsaac:


N1x0:
Is this just future proofing or is `SearchHost` present and used in windows 10 as well?

dalyIsaac:
It's not used in the stable Windows 10 builds, as far as I know.

N1x0 (approved):
Might be a while until this becomes relevant, but better to cover it now since there are no adverse effects I can think of.

josteink (approved):

Issue #110, opened 2020-07-25
Using the Win key in keybindings that focus windows causes the start menu to appear


TriCrose:
I'd like to use the Win key for my keybindings (which works fine for the most part), but I've noticed that any actions that involve focusing a window will end up showing the start menu immediately, thereby taking the focus away from the window. E.g. try assigning `FocusNextWindow()`/`FocusPreviousWindow()` to Win + Left/Right and you'll see this issue.

TriCrose:
Having looked through the code a bit, I think the issue is happening on [this line](https://github.com/rickbutton/workspacer/blob/master/src/workspacer.Native/Native/FocusStealer.cs#L19) in `FocusStealer`:
```C++
Win32.keybd_event(0, 0, 0, UIntPtr.Zero);
```
As I've found that commenting it out stops the start menu from triggering, however it also prevents the `FocusStealer` from actually stealing the focus. The commit message for this line is "switched to new keybd_event hack for FocusStealer", perhaps this is a side-effect of that?

imswebra:
See [my comment](https://github.com/workspacer/workspacer/pull/231#issuecomment-873128697) on #231 for a current workaround using [Open-Shell](https://github.com/Open-Shell/Open-Shell-Menu).

system32uwu:
Should we add a "Troubleshooting" section including this? There's a way to disable Win + L but It won't let you lock Windows ever again unless you edit the registry.

adolfocorreia:
I came up with another workaround for this using AutoHotKey. In the workspacer configuration, I map the desired action to some alternative key binding (e.g. Win+Alt+J) and then in AHK I bind the desired Win key binding (e.g. Win+J) to the other one. It's not completely flawless, but it gets the job done. In my experience, AHK usually handles Win key bindings better than alternative software. This hack won't work for all keys though (e.g. Win+L).

Here is some example code:
```cs
KeyModifiers modA  = KeyModifiers.Win | KeyModifiers.Alt;
k.Subscribe(modA, Keys.J, () => w.FocusedWorkspace.FocusNextWindow(),     "focus next window");
k.Subscribe(modA, Keys.K, () => w.FocusedWorkspace.FocusPreviousWindow(), "focus previous window");
```
AHK:
```ahk
#j::SendInput #!j
#k::SendInput #!k
```

Issue #195, opened 2021-03-10
Changing mod key to LWin/Win


jdscanlon:
I have zero experience in C#, and I cannot figure out how to rebind Win/super to mod. The only way that I can have it function at all is to bind keys individually. It seems to be bound to alt permanently unless I do _'UnsubscribeAll'_. Anyway, If there is a one-liner to fix this, or someone else's config using super, I'd really appreciate it.  I am using the latest release, 0.9.10. I would post my config, but embarrassingly I've deleted it about 30 times now trying to figure this out.

You guys are doing some great work here. I'd like to submit some documentation for this project once I get my head around the basics. Thanks!

HiMyNameIsIlNano:
@jdscanlon wouldn't it be enough to change this `KeyModifiers mod = KeyModifiers.Alt;` from the default configuration to the Win key? I am not an active user of this project (yet) so please let me know if you already tried that. 

Cheers,
HiMyNameIsIlNano

jdscanlon:
You would think so, but I spent an embarrassingly long time attempting this, plus many other things. I have yet to find any method of doing so, besides rebinding all keys. 

Unfortunately I am becoming pretty comfy on alt at this point, and it's messing with my head on other OS's lol.

dalyIsaac:
There's a method [`SubscribeDefaults`](https://github.com/rickbutton/workspacer/blob/32783981f22ed2bbf7a32565888b3a27dcf923e7/src/workspacer/Keybinds/KeybindManager.cs#L223) which should let you do it. 

https://github.com/rickbutton/workspacer/blob/32783981f22ed2bbf7a32565888b3a27dcf923e7/src/workspacer/Keybinds/KeybindManager.cs#L223

However, it's not exposed as part of the [`IKeybindManager`](https://github.com/rickbutton/workspacer/blob/32783981f22ed2bbf7a32565888b3a27dcf923e7/src/workspacer.Shared/Keybind/IKeybindManager.cs), so I don't think it's possible to use in your config - it might require a PR.

N1x0:
Just wanted to add that remapping to `Win` can be difficult and introduce weird behaviour (`Win + L` is reserved for locking the screen for instance and this can only be changed through a registry hack, similar changes are required for all other reserved key combinations such as `Win + E` , or `Win  + F`.

[...]

PR #231, opened 2021-07-01
Reverted to AllocConsole method for focus

system32uwu:
See #110, I'm having this exact problem using the Win key, the start menu appears when I switch workspaces with Win + {1..9} (which is annoying of course). Not sure why @rickbutton made this change when this was working properly.

### DIFF
--- a/src/workspacer.Native/Native/FocusStealer.cs
+++ b/src/workspacer.Native/Native/FocusStealer.cs
@@ -16,8 +16,11 @@ namespace workspacer
 
         public static void Steal(IntPtr windowToFocus)
         {
-            Win32.keybd_event(0, 0, 0, UIntPtr.Zero);
-
+            Win32.AllocConsole();
+            var hWndConsole = Win32.GetConsoleWindow();
+            Win32.SetWindowPos(hWndConsole, IntPtr.Zero, 0, 0, 0, 0, Win32.SetWindowPosFlags.IgnoreZOrder);
+            Win32.FreeConsole();
+            
             Win32.SetForegroundWindow(windowToFocus);
         }
     }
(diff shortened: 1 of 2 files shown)


dalyIsaac:
Isn't this related to #195 too?

system32uwu:
> I personally think it'd be better to recommend using modifiers like `Win + Shift` instead of using `Win` by itself.

The thing is most users (like me) come from other unix tiling window managers like xmonad, dwm or i3, and we all usually use the windows key as the mod key. So it will be very inconvenient having to use a different mod key when you switch OS (I dual boot windows for dotnet and linux as my main driver).

Also, the Win key itself has no useful functionalities behind other than opening the start menu. In case this is still unconvincing we should either: mention that the use of the win key is not recommended due to this problem, or do something to prevent the start menu from opening. 

This seems to be the easiest solution to this problem, but if there's a better way to do it I have no problem

imswebra:
Strongly echo the importance for the Windows key to be (properly) bindable for keybind parity with Unix tiling windows managers. I imagine there is a fairly significant chunk of the userbase here that use workspacer because they are coming from a Unix system with a proper tiling wm and want their workflow to stay the same.

As a current stopgap, if you can live with not being able to open the start menu with the Windows key, I have found that I can use [Open-Shell](https://github.com/Open-Shell/Open-Shell-Menu) to disable the start menu popping up when using keybindings with the Windows key (#110).
Under the "Controls" tab of the Open-Shell settings, I just set Win and Shift+Win to do nothing (and then keep left click to open the W10 start menu so it's still accessible). The lack of start menu is eased if you use a dmenu/Rofi style launcher like [Wox](https://www.wox.one) or [Keypirinha](https://keypirinha.com), with my preference being the latter.

Note however that this does not solve conflicts with some Windows shortcuts. I can map Win+F just fine (typically a Window feedback shortcut) for example, but to use Win+L for vim-style navigation keys, I had to use a registry hack to disable the typical lock shortcut.

system32uwu:
Thank you @imswebra, I chose "Nothing" instead:
image. I'll be working on adding a workaround to this in the docs with the information you provided. I'm closing this PR.

PR #233, opened 2021-07-04
Prevent switching to newly added windows

### LINKED ISSUE #159: Monitors switch between workspaces when opening application
Sometimes when i open an application my 2 monitors switch workspaces, for example: `monitor 1` is on workspace `3` and `monitor 2` is on workspace`9`, I open an application and `monitor 1` is on workspace `9` and `monitor 2` is on workspace `3`.

I think this is a problem with the monitor array switching order which makes `monitor 1` `monitor 2` ext. but it's just speculation.

let me know if you need more data, I know this data isn't very useful unless you also have this problem but I don't really have more right now.

dalyIsaac:
Resolves #159 

The only change is preventing `AddWindow` to switching to the window:

https://github.com/workspacer/workspacer/blob/7f2dc71474037667c986b1b1df7d3c5e70eb80d7/src/workspacer/Workspace/WorkspaceManager.cs#L347-L352

All the other changes seem to be due to a mismatch between line endings (sorry).


system32uwu (approved):

josteink:
This PR seems to co flare two things (Win11-fix and real bug fix) but otherwise looks ok to me.

josteink (approved):

PR #234, opened 2021-07-05
New Hugo theme

BeeryShklar:
I've made a new theme for the website. What do you think?
Please tell me if there's something you want to change.

# Screenshots

## Home Page
Home Page

## Configuration Page
Config Page

## Light Version
There's also a light version which is chosen by the user's system prefrences
Light Version

## 404 Page
404 Page

## Responsive Header
Responsive Header


N1x0:
I will need some time to review all the changes but design-wise I love it! Might make sense to put in some more work into the contents before pushing this. Maybe we could even set up a seperate milestone.

BeeryShklar:
@N1x0 I agree about adding some more content. The website really doesn't have much content currently.

docs/themes/workspacer/static/css/style.css
```diff
@@ -0,0 +1,243 @@
...
+	padding: 1rem;
+	border-radius: 0.5rem;
+	overflow-x: auto;
+}
+
+header, footer {
+	position: relative;
+	margin: 0 1rem;
+	height: 8vh;
+	padding: 2vh 1rem;
```
dalyIsaac, line 93: On large monitors using `2vh` causes the header too large

dalyIsaac (changes requested):
Thanks for the PR @BeeryShklar! Just a couple of things:

1. Could font sizes be set in `px`? A bunch of different selectors specify `font-size` using relative units like `rem`.

2. It'd also be great if `code` sections could be distinguished better - currently the colors (at least in the dark theme) are a bit similar:

image

3. What was the rationale behind moving the website root to using a horizontal layout? It's slightly inconsistent with the other pages.

BeeryShklar:
@dalyIsaac 
1. It is considered best practice to use rem over px. rem is essentially 16px but if you change the root's font size all the elements using rem will also change their size.
2. Changing the `code` bg color is difficult because it is hard coded into the element but I will try to change it.
3. The horizontal layout is suppose to mimic the tiling aspect of Workspacer. Under about 1000px of viewport width it goes back to vertical.
4. I will try to have the header at a fixed size.

dalyIsaac:
> 1. It is considered best practice to use rem over px. rem is essentially 16px but if you change the root's font size all the elements using rem will also change their size.

👍 

> 2. I changed the `pre` color to a lighter color than the bg. I had to use `!important` but it works.

Thanks, looks much better now.

> 3. The horizontal layout is suppose to mimic the tiling aspect of Workspacer. Under about 1000px of viewport width it goes back to vertical.

Makes sense. There's a couple of issues with small window sizes:

image

image

> 4. I changed the `header` height to 5rem.

Thanks!

dalyIsaac (approved):
LGTM!

docs/config.toml
```diff
@@ -37,3 +37,20 @@ publishDir = "../out/docs/"
 [[menu.extra]]
     name = "submit an issue"
     url = "https://github.com/workspacer/workspacer/issues/new"
+
+[[menu.footer]]
+    name = "Workspacer"
+    url = "https://github.com/workspacer/workspacer"
+    weight = 1
+    
+[[menu.footer]]
```
system32uwu, line 46: Since workspacer is now a community project, shouldn't we remove this?

docs/config.toml
```diff
@@ -37,3 +37,20 @@ publishDir = "../out/docs/"
...
+    name = "Rick Button"
+    url = "https://github.com/rickbutton"
+    weight = 2
+    
+    
+[[menu.footer]]
```
system32uwu, line 52: The same with this, also the website is down.

docs/config.toml
system32uwu: (line 46 - 49) not sure why it's displayed like this

docs/config.toml
system32uwu: (line 52 - 55)

docs/config.toml
BeeryShklar: Maybe, but I think Rick deserves some credit for his work.

docs/config.toml
```diff
@@ -37,3 +37,20 @@ publishDir = "../out/docs/"
...
+    name = "Website"
+    url = "https://rickbutton.me"
```
system32uwu, line 54: Apparently his new website is https://button.dev/, in case we want to keep this reference this should be updated

josteink:
We've had some stoppage in PR-processing due to a 2-review requirement. This has now been reduced to 1, so consider this merged. Sorry about the delay!

PR #247, opened 2021-07-26
Add winget installation

sitiom:
Edit: It was actually added to winget way back last year but it came off the top of my head (winget wasn't still stable at that time).

### DIFF
--- a/docs/content/quickstart.md
+++ b/docs/content/quickstart.md
@@ -8,20 +8,25 @@ type: docs
 
 Download the installer or the zip file from the [Releases](https://github.com/workspacer/workspacer/releases/latest) page and install/unzip it anywhere you like. 
 
-Alternatively, workspacer is available via [Chocolatey](https://community.chocolatey.org/packages/workspacer):
+Alternatively, workspacer is available via these package managers:
 
+[WinGet](https://winstall.app/apps/rickbutton.workspacer)
+```shell
+winget install workspacer
+```
+
+[Chocolatey](https://community.chocolatey.org/packages/workspacer)
 ```shell
 choco install workspacer
 ```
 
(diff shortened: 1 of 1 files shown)


system32uwu:
Looks like it was uploaded by [rickbutton](https://winget.run/pkg/rickbutton/workspacer), has anyone tried installing through winget? Not sure how it works, but the version there is the latest stable release, isn't there a way to also get versions from other tags? (like beta or unstable). This is fine btw, but since that release is from last year I don't know how many people would want to use this method.

system32uwu (approved):

sitiom:
> Looks like it was uploaded by [rickbutton](https://winget.run/pkg/rickbutton/workspacer),

Actually, I do remember now. [It was me!](https://github.com/microsoft/winget-pkgs/pull/5731) (Not sure how I forgot that I myself put it there. I use chocolatey most of the time, and there was a package fix just yesterday and I thought it was just recently added 😅)

> has anyone tried installing through winget? Not sure how it works, but the version there is the latest stable release, isn't there a way to also get versions from other tags? (like beta or unstable). This is fine btw, but since that release is from last year I don't know how many people would want to use this method.

Yes, it does work. But support for different release channels is still an ongoing issue (https://github.com/microsoft/winget-cli/issues/147). Would need to wait for that.

dalyIsaac:
Speaking of releases, it'd be great if #233 could be reviewed - once that's done I'll update the beta branch.

dalyIsaac (approved):

sitiom:
> Speaking of releases, it'd be great if #233 could be reviewed - once that's done I'll update the beta branch.

Also would need to fix this as well https://github.com/workspacer/workspacer/issues/248.

PR #249, opened 2021-07-26
Fix autoupdater and add beta links

### LINKED ISSUE #248: Autoupdater not working
Would need to fix the file format with the new one I changed in #236.
https://github.com/workspacer/workspacer/blob/a89d3449fc969053a65e849d0f9ce5d7f2ec0a8b/src/workspacer/Program.cs#L92

Also, there is no mechanism to get the link for the beta branch. Would fix that as well.

sitiom:
Fixes #248. Added the beta links.

Updated the WiX XML to set the `InstallLocation` Regkey when installing. Now, it can differentiate and download the appropriate file whether installed in the system (MSI) or portable (zip).


dalyIsaac (approved):
LGTM!

josteink (approved):
LGTM.

PR #246, opened 2021-07-24
Allow window title bars to be removed

dalyIsaac:
Resolves #217 

This adds the ability to hide window title bars, as a new plugin. This only applies to windows created _after_ workspacer starts up.

**Changes:**

- Added `workspacer.TitleBar` project.
- Changed `Win32.WS` and `Win32.WS_EX` to be `long` instead of `ulong`.
  I did this because Win32 functions like [`SetWindowLongPtr`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptrw) specify [`LONG_PTR`](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/bbcfb0af-8349-4e98-ad26-957e1363f714) instead of [`ULONG_PTR`](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/21eec394-630d-49ed-8b4a-ab74a1614611) (though I'm not sure it matters much).
- Added methods `SetWindowStylePtr` and `SetWindowStyleExLongPtr` to set window styles.
- Add the ability to set/remove `WS_CAPTION` and `WS_THICKFRAME` (see [docs.microsoft.com: Window Styles](https://docs.microsoft.com/en-us/windows/win32/winmsg/window-styles)).
- Added documentation to `config.md`

**Available window styles:**

image

(NOTE: screenshot taken on Windows 11, hence the rounded borders)

PR #253, opened 2021-08-04
Seperate source and test projects

sitiom:
My point stands at https://github.com/workspacer/workspacer/pull/246#issuecomment-892262982.


josteink:
Following up on discussion from previous PR here.

> It's not a less important folder, it keeps things organized 

And it's IMO not really "organizing" anything by putting tests in a separate folder. You're just creating a code-bin which is easily collapsed and/or ignored.

Organizing something means actually creating some sort of organization. IMO creating a separate tests-folder always (without exception) results in the opposite. The "tests" folder is now the entirety of the "organization". And that's the end of that. It's really a absolutely lowest aim of organization one can have.

In my experience having tests side-by-side with the projects being tested has the following benefits:

* You always know where to find tests for the project you're working with (in the test-project below the one you're working on).
* You always know where to put new tests when writing new code (in the test-project below the one you're working on).
* You always know if a project has tests or not (is there a test-project beneat it in solution explorer?), and if not, where to create the test-project and with what name (project-name + `.Tests`), because that's the way you are (naturally) forced to *organize* it.
* Everything is organized, always, because it feels *natural* to keep things organized.

Contrast that with having a separate test-folder:

* There's no natural forces driving you to create a corresponding test-project for the project being tested.
* There's no immediate way to know where to put tests for your project
* There's no immediate way to tell if your project has tests or not
* There's no easy way to spot projects which stand out by lacking tests
* Etc etc
* Easy to minimize/collapse.. Ignore. Don't keep up.

There are IMO no objective benefits to having a separate folder for test-projects, only downsides.

> that's how most projects actually do it.

Just because other projects do it that way doesn't automatically make it better. Doing things just because other's are doing it, is  known as [cargo-cult computing](https://en.wikipedia.org/wiki/Cargo_cult_programming).

And that's why most projects do it. Cargo cult-programming or just "default" conventions, with no real *reflection* behind the purpose of doing things in the way they've chosen.

Here on the contrary, I've put lots of thought and reflection into *not* having a separate tests-folder, and I would appreciate if we didn't change that.

josteink (changes requested):
My vote is that we do *NOT* commit any change like this.

dalyIsaac (changes requested):
I agree with @josteink in this case, especially with the following points:

> - There's no natural forces driving you to create a corresponding test-project for the project being tested.
> - There's no immediate way to know where to put tests for your project
> - There's no immediate way to tell if your project has tests or not
> - There's no easy way to spot projects which stand out by lacking tests
> - Easy to minimize/collapse.. Ignore. Don't keep up. And wait for the code-rot to take hold.

As it is, this project doesn't have a lot of tests, and when browsing the repository, as far as I can tell @josteink authored the vast majority of them. 

I will admit that I accidentally had a commit which moved tests to the wrong location, which was caused by Visual Studio hiding the directory structure. However, I immediately picked this up when I switched to vscode, which shows the folder structure. I would definitely fall victim to hiding the `test` folder.

sitiom:
I see the point now. Maybe it is hard to keep the `src` and `test` folders parallel. The idea of everything being into "one folder" seems like clutter to me but that doesn't seem to matter.

PR #264, opened 2021-10-11
Show all window titles in TitleWidget

denjukebox:
The PR expands the functionality of the title widget so all window titles in the workspace are mentioned. This is needed for stacked/tabbed layouts.

- Part with per window in workspace
- Highlighting on selected window
- Click action on window to bring to front
- Max length for titles

In master there is an issue with BarSection where previous created parts are still shown (and active) on title bar without being returned after refresh. This is solved in https://github.com/workspacer/workspacer/pull/268


src/workspacer.Bar/Widgets/TitleWidget.cs
```diff
@@ -1,49 +1,57 @@
...
-            Context.Workspaces.FocusedMonitorUpdated += RefreshFocusedMonitor;
+            if (currentWorkspace.LastFocusedWindow != null && allWindows.Contains(currentWorkspace.LastFocusedWindow) &&
+                currentWorkspace.FocusedWindow != currentWorkspace.LastFocusedWindow)
+            {
+                //Move to second place
+                allWindows.Remove(currentWorkspace.LastFocusedWindow);
+                allWindows.Insert(1, currentWorkspace.LastFocusedWindow);
+            }
+
+            return allWindows.Where(window => !filterOnTitleFilled || !string.IsNullOrEmpty(window.Title)).ToArray();
```
dalyIsaac, line 54: `Where` returns an `IEnumerable` which fulfils `GetWindows`' return type, so `ToArray` isn't necessary.

src/workspacer.Bar/Widgets/TitleWidget.cs
```diff
@@ -1,49 +1,57 @@
...
 namespace workspacer.Bar.Widgets
 {
     public class TitleWidget : BarWidgetBase
     {
-        public Color MonitorHasFocusColor { get; set; } = Color.Yellow;
+        #region Properties
+        public Color WindowHasFocusColor { get; set; } = Color.Yellow;
         public bool IsShortTitle { get; set; } = false;
+        public int? MaxTitleLength { get; set; } = null;
+        public bool MultiWindowTitles { get; set; } = false;
```
dalyIsaac, line 13: Can this be renamed to `ShowAllWindowTitles`?

dalyIsaac (changes requested):
I tried it out - nice work 🎉 
I was expecting it to behave like `WorkspaceWidget` where the window list would stay static. 

Can you also update the documentation in `docs/` with an example?

src/workspacer.Bar/Widgets/TitleWidget.cs
```diff
@@ -1,51 +1,38 @@
...
-                    return Parts(Part(shortTitle, color, fontname: FontName));
-                }
-            }
-            else
-            {
-                return Parts(Part(NoWindowMessage, color, fontname: FontName));
-            }
-        }
+        public Func<IWindow, Action> TitlePartClicked = ClickAction;
+        public Func<IWindow, object> OrderWindowsBy = (window) => { return $"{window.IsFocused}{window.Title}"; };
```
dalyIsaac, line 18: I'm not sure what other WMs do, but to me, as this has some resemblance to `WorkspaceWidget`, I would (perhaps naively) expect the ordering of the window titles to reflect the current layout's stack and stay static.

```suggestion
        public Func<IWindow, object> OrderWindowsBy = (window) => 0;
```

denjukebox: I agree that the default of unsorted is better and is the more default. Allowing it to be customized i forgot to set it to static.

dalyIsaac (approved):
LGTM. I'll wait for #268 before merging this.

dalyIsaac:
After some further testing, I found that swapping window locations (e.g., using `workspaces.FocusedWorkspace.SwapFocusAndNextWindow()`) doesn't trigger an update of the order of the items in `TitleWidget`. This is because `Workspace.SwapWindows` doesn't trigger any events in `WorkspaceManager`.

One workaround would be triggering the `Workspace.WindowMoved` event from a `Workspace` instance. However, I'm not convinced this is the best move.

Do you have any thoughts?

josteink:
@dalyIsaac : You seem to request more action on the submitter, yet the PR is approved.

Should we merge or not? :smile:

dalyIsaac:
Sorry for the confusion. As it stands, I don't think we should merge.

denjukebox:
It took me a while to get back to this but after reading the code a bit more in Workspace & WorkspaceManager I do have some input. 

First of I understand the issue but think it should be addressed under a bug and/or different PR to discuss the issue as it is one inherent to the way the aforementioned classes are written and not this PR's modification to TitleWidget.

So now for my suggestion, I would move the handling of Window changes to the Manager class and adjust those functions to be able to receive the specific workspace on which the action needs to happen. This way the Workspace class is limited in scope to containing the data and not the logic of manipulation. That way all events can be triggered from the same instance.

denjukebox:
After a hiatus I've fixed the issue where switching windows was not triggering an update of the titlewidget with the help of @josteink under the draft that was mentioned before. As that was the blocker for this PR I would like to further pick this one up and complete the functionality.

josteink:
CC: @tlsomers 

This PR has been dormant for a while. Could you take a look at it, give it a review, and merge if you think it looks good?

src/workspacer.Bar.Tests/TitleWidgetTests.cs
```diff
@@ -15,7 +15,7 @@ public void GetShortTitle_SplitHyphen()
         [Fact]
         public void GetShortTitle_SplitEmDash()
         {
-            var s = "workspacer � Visual Studio Code";
+            var s = "workspacer — Visual Studio Code";
```
josteink, line 18: Is this change intentional? Or an encoding bug? :smile:

src/workspacer.Bar/Widgets/TitleWidget.cs
```diff
@@ -5,44 +5,33 @@ namespace workspacer.Bar.Widgets
...
-            Context.Workspaces.WindowAdded += RefreshAddRemove;
-            Context.Workspaces.WindowRemoved += RefreshAddRemove;
+            Context.Workspaces.WindowAdded += RefreshAdd;
+            Context.Workspaces.WindowRemoved += RefreshRemove;
             Context.Workspaces.WindowUpdated += RefreshUpdated;
             Context.Workspaces.FocusedMonitorUpdated += RefreshFocusedMonitor;
         }
 
+        #region Get Windows
+        private IEnumerable<IWindow> GetWindows(bool filterOnTitleFilled = true)
```
josteink, line 29: `IEnumerable<T>` requires you to add a using statement on top for `System.Collections.Generic` :smile:

josteink: I believe this is the singular cause of the build failure and we won't be able to merge until this is resolved.

denjukebox:
@josteink Thank you for resolving these last stupid issues!

josteink:
What can I say? I hate seeing perfectly good PRs go to waste :smile:

PR #267, opened 2021-10-18
Changes LoadFile to LoadFrom to load DLLs included by plugins

denjukebox:
This PR fixes an issue that was encountered trying to chain load DLL's included by a plugin.

Developing a plugin for sound control required an external DLL that refused to be found by Assembly.LoadFile. Upon further investigation replacing the function to LoadFrom fixes this without side-effects to existing functionality.


src/workspacer/Plugins/PluginManager.cs
```diff
@@ -52,8 +50,7 @@ private static Type GetPluginFromFolder(string dir)
             var name = new DirectoryInfo(dir).Name;
 
             Console.WriteLine(Path.Combine(dir, name + ".dll"));
-            var assembly = Assembly.LoadFile(Path.Combine(dir, name + ".dll"));
-
+            var assembly = Assembly.LoadFrom(Path.Combine(dir, name + ".dll"));
```
josteink, line 53: The absolutely best thing to do would be to unit-test this, but I can see how creating a test-case for this is going to be fairly involved.

As a second best option, if the use of `LoadFrom()` is critical for correct operations., maybe we should add a comment saying that much wherever we use it?

denjukebox: I would also opt for the second option. I'll write up a comment referring to https://docs.microsoft.com/en-us/archive/blogs/suzcook/loadfile-vs-loadfrom as it explains the situation rather well. There is a third option of binding to the AssemblyResolve event but that overcomplicates things for the project's scope.

josteink (approved):

josteink:
Thanks for the PR. Consider it merged!

PR #268, opened 2021-10-18
Rework of Barsection to solve child cleanup

denjukebox:
Reworked creation of parts for BarSection for proper cleanup of unused childs.
Widgets each have seperate FlowLayout within Barsection.

Corrected some instances of config not being passed fully.


dalyIsaac:
I'm sorry, could you provide some more details for what this PR is trying to solve? Additionally, why do widgets within `BarSection` now have their own `FlowLayoutPanel`?

denjukebox:
When developing the titlewidget i noticed that not all controls where getting cleaned up when i was returning less parts. Doing it this way solved that. The flowpanels make it easy to manage the creations and removals. It feels logical to me that each plugin would have its own manageable space on the bar.

With the separate flowlayoutpanels you can allow for better plugin specific adjustment of how the controls are shown within its section in the bar. Centring, width, margin being some of them. It also allows changing what controls they can return. If you put the create and update logic of the controls in the Part classes and because of this change all controls in the flowpanel are the same, [complex controls](https://github.com/denjukebox/workspacer/commits/feat/BarPlugin_Complex_Controls) becomes possible. I've implemented this in titlewidget where the process icon is returned with the text. 
afbeelding

src/workspacer.Bar/BarSection.cs
```diff
@@ -38,85 +38,95 @@ public class BarSection
...
+                    AddLabel(pannel);
+                }
+
+                while (pannel.Controls.Count > partCount)
                 {
-                    for (var i = partNumber; i < _panel.Controls.Count; i++)
-                    {
-                        toRemove.Add(_panel.Controls[i]);
-                    }
+                    pannel.Controls.RemoveAt(0);
```
dalyIsaac, line 99: Wouldn't it be more efficient to have the removals occur at the end of the collection?

dalyIsaac (changes requested):
Ah I see. Thanks for bearing with me there. I think with a few minor changes it'll be good to go.

denjukebox:
Thanks for taking the time to read my changes and catching my faults. I corrected them and realised while reading the suggested fixes for https://github.com/workspacer/workspacer/pull/272 that with this change we have the ability to only update the widgets that mark themself's dirty, allowing the timers already present to not interfere with each other in needing refreshes. I moved the dirty logic to the widget interface and adjusted the Draw function accordingly. Should optimise things a bit.

dalyIsaac (approved):
LGTM

Issue #265, opened 2021-10-11
CPU/memory indicator widget


adolfocorreia:
Here is some code I came up with to have a CPU/memory consumption indicator bar widget. Feel free to incorporate it into workspacer if you wish.

```csharp
using System.Diagnostics;
using Microsoft.VisualBasic.Devices;
// (other usual imports)
class CpuMemWidget : BarWidgetBase {
    private double _interval;
    private string _template;
    private ulong _totalMemory;
    private Timer _timer;
    private PerformanceCounter _cpuCounter;
    private PerformanceCounter _memCounter;
    private string _text;
    public CpuMemWidget(double interval, string template) {
        _interval = interval;
        _template = template;
    }
    public override void Initialize() {
        _totalMemory = new ComputerInfo().TotalPhysicalMemory / (1024*1024);
        _cpuCounter = new PerformanceCounter("Processor", "% Processor Time", "_Total", true);
        _memCounter = new PerformanceCounter("Memory", "Available MBytes", true);
        _text = _GenerateTextFromCounters();
        _timer = new Timer(_interval);
        _timer.Elapsed += (s, e) => {
            _text = _GenerateTextFromCounters();
            Context.MarkDirty();
        };
        _timer.Enabled = true;
    }
    public override IBarWidgetPart[] GetParts() {
        return Parts(Part(_text, fontname: FontName));
    }
    private string _GenerateTextFromCounters() {
        string cpu = Convert.ToInt32(_cpuCounter.NextValue()).ToString();
        string mem = Convert.ToInt32((1 - _memCounter.NextValue()/_totalMemory) * 100).ToString();
        return _template.Replace("[cpu]", cpu).Replace("[mem]", mem);
    }
}
```
Usage example, with 15 seconds refresh and Unicode symbols:
```csharp
new CpuMemWidget(1000 * 15, " \uf2db [cpu]%  \uf538 [mem]%")
```

josteink:
That looks nice-ish and might be a good default widget actually.

Care to create a PR for this? :)

denjukebox:
I took the liberty of implementing your idea into a PR that included some more functionality. You should be credited in the PR.

adolfocorreia:
Thanks for creating the PR, @denjukebox! From what I could see, your implementation is more general than mine.

josteink:
If it's OK with you @adolfocorreia , I guess that means we can/should close this PR?

adolfocorreia:
[...]

PR #272, opened 2021-11-03
BarPlugin PerformanceWidget

denjukebox:
This PR implements 3 performance counter based widgets to display statistics about the system in the bar.

- Extendable PerformanceCounter base class
- CPU Widget
- Memory Widget
- Network Widget

The original idea was suggested by @adolfocorreia
https://github.com/workspacer/workspacer/issues/265


src/workspacer.Bar/Widgets/PerformanceWidget.cs
```diff
@@ -0,0 +1,163 @@
...
+            var interfaces = category.GetInstanceNames();
+            if (!interfaces.Contains(interfaceName))
+            {
+                _interfaceName = interfaces.First();
+            }
+        }
+
+        protected override string GetText(float value)
+        {
+            var amount = value;
```
dalyIsaac, line 136: I don't think the `amount` variable is necessary.

src/workspacer.Bar/Widgets/PerformanceWidget.cs
```diff
@@ -0,0 +1,163 @@
...
+            return Parts(Part($"{GetText(_counter.NextValue())}{icon}", null, null, ClickAction));
+        }
+
+        private string GetIconOrEmpty()
+        {
+            var icon = string.Empty;
+
+            try
+            {
+                icon = char.ConvertFromUtf32(int.Parse(Icon, System.Globalization.NumberStyles.HexNumber));
```
dalyIsaac, line 66: I think it'd be nice if we could allow custom text and emojis.

src/workspacer.Bar/Widgets/PerformanceWidget.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    public class NetworkPerformanceWidget : PerformanceWidgetBase
+    {
+        protected override string CounterCategory => "Network Interface";
+        protected override string CounterName => "Bytes Total/sec";
+        protected override string CounterInstance => _interfaceName;
+        protected override string Icon => "1F5A7";
+        protected override Action ClickAction => () => Process.Start("explorer.exe", @"shell:::{26EE0668-A00A-44D7-9371-BEB064C98683}\3");
```
dalyIsaac, line 119: Can you add a comment describing what this points to?

dalyIsaac (changes requested):
This looks awesome!

Can you also provide some documentation in `/docs` for potential usage?

adolfocorreia:
Great work on this new PerformanceWidget, @denjukebox!

I was not able to run/test your code, but if I may make a comment on this PR, I would suggest against calling `_counter.NextValue()` in the `GetParts()` function and poll the counter only in the timer callback instead. That was an issue I had in my implementation. The problem was that the CPU/memory widget might be redrawn because of other events (e.g. a TimeWidget update), and a new counter poll might be run before the next timer event. You could try to reproduce this behavior by setting the PerformanceWidget with a long update interval (e.g. 1 minute) and a TimeWidget with a short update interval (e.g. 1 second).

dalyIsaac:
Thanks for adding the documentation. I think it'd be good to update based on @adolfocorreia's feedback.

denjukebox:
@adolfocorreia Thanks for noticing the update issue! I'd seen the numbers updating a bit quickly put didn't put things together.

adolfocorreia:
Glad I could help!

But I think it is still necessary to call `Context.MarkDirty()` in the timer callback, isn't it? Otherwise the PerformanceWidget might not be updated on the screen if there aren't other redraw events.

dalyIsaac:
> Glad I could help!
> 
> But I think it is still necessary to call `Context.MarkDirty()` in the timer callback, isn't it? Otherwise the PerformanceWidget might not be updated on the screen if there aren't other redraw events.

@denjukebox I think this might be the case.

denjukebox:
Yup totally missed that. I added it in again with an overridable update function for the timer.

dalyIsaac:
For `NetworkPerformanceWidget`, I think it'd be worth adding that to select the appropriate instance, you can get the name from Device Manager:

image

would be instantiated as:

```csharp
new NetworkPerformanceWidget("Intel[R] Ethernet Connection [7] I219-V"),
```

xd999e:
Any plans to merge this feature into main?

denjukebox:
Apparently this was still blocked by @dalyIsaac his request for documentation. I've done this in the mean while but I guess this has to be checked off still? Been away from this project for a bit now.

dalyIsaac (approved):

denjukebox:
@dalyIsaac  Thanks for the merge an grammar fixes!

pranavpa8788:
Whenever I use any of the performance widgets I get the The type or namespace name 'NetworkPerformanceWidget' could not be found (are you missing a using directive or an assembly reference?). 

I am using the latest verison of workspacer. Do I have to manually compile the dll file?

PR #274, opened 2021-11-19
Workspace events on swap and move of windows

denjukebox:
Moved window manipulation methods to workspace manager and added in events to communicate window movements to subscribers. 

This is a fix PR that requests input from maintainers to verify method of working is wanted

### DIFF
--- a/src/workspacer.Native/Native/WindowsManager.cs
+++ b/src/workspacer.Native/Native/WindowsManager.cs
@@ -14,7 +14,7 @@ using System.Windows.Forms;
 namespace workspacer
 {
     public delegate void WindowDelegate(IWindow window);
-    public delegate void WindowAddDelegate(IWindow window, bool firstCreate);
+    public delegate void WindowCreateDelegate(IWindow window, bool firstCreate);
     public delegate void WindowUpdateDelegate(IWindow window, WindowUpdateType type);
 
     public class WindowsManager : IWindowsManager
@@ -30,13 +30,13 @@ namespace workspacer
 
         private Dictionary<WindowsWindow, bool> _floating;
 
-        public event WindowAddDelegate AddWindowTrigger;
-        public event WindowDelegate RemoveWindowTrigger;
-        public event WindowUpdateDelegate UpdateWindowTrigger;
+        public event WindowCreateDelegate WindowCreated;
+        public event WindowDelegate WindowDestroyed;
+        public event WindowUpdateDelegate WindowUpdated;
 
-        public event WindowFocusDelegate WindowFocused;
-        public event WindowDelegate WindowUpdated;
-        public event WindowDelegate WindowClosed;
+        public event WindowFocusDelegate OnWindowFocused;
+        public event WindowDelegate OnWindowUpdated;
+        public event WindowDelegate OnWindowClosed;
 
         public IEnumerable<IWindow> Windows => _windows.Values;
 
@@ -220,9 +220,9 @@ namespace workspacer
 
                 if (!ShouldIgnoreWindow(window))
                 {
-                    window.WindowFocused += () => HandleWindowFocused(window);
-                    window.WindowUpdated += () => HandleWindowUpdated(window);
-                    window.WindowClosed += () => HandleWindowClosed(window);
+                    window.WindowFocused += (sender) => HandleWindowFocused(sender);
+                    window.WindowUpdated += (sender) => HandleWindowUpdated(sender);
+                    window.WindowClosed += (sender) => HandleWindowClosed(sender);
 
                     _windows[handle] = window;
 
@@ -249,7 +249,7 @@ namespace workspacer
             if (type == WindowUpdateType.Show  && _windows.ContainsKey(handle))
             {
                 var window = _windows[handle];
-                UpdateWindowTrigger?.Invoke(window, type);
+                WindowUpdated?.Invoke(window, type);
             }
             else if (type == WindowUpdateType.Show)
             {
@@ -263,13 +263,13 @@ namespace workspacer
                     UnregisterWindow(handle);
                 } else
                 {
-                    UpdateWindowTrigger?.Invoke(window, type);
+                    WindowUpdated?.Invoke(window, type);
                 }
             }
             else if (_windows.ContainsKey(handle))
             {
                 var window = _windows[handle];
-                UpdateWindowTrigger?.Invoke(window, type);
+                WindowUpdated?.Invoke(window, type);
             }
         }
 
@@ -281,7 +281,7 @@ namespace workspacer
                 Logger.Debug("StartWindowMove[{0}]", window);
 
                 HandleWindowMoveStart(window);
-                UpdateWindowTrigger?.Invoke(window, WindowUpdateType.MoveStart);
+                WindowUpdated?.Invoke(window, WindowUpdateType.MoveStart);
             }
         }
 
@@ -293,7 +293,7 @@ namespace workspacer
                 Logger.Debug("EndWindowMove[{0}]", window);
 
                 HandleWindowMoveEnd();
-                UpdateWindowTrigger?.Invoke(window, WindowUpdateType.MoveEnd);
+                WindowUpdated?.Invoke(window, WindowUpdateType.MoveEnd);
             }
         }
 
@@ -304,24 +304,24 @@ namespace workspacer
                 var window = _windows[handle];
                 if (_mouseMoveWindow == window)
                 {
-                    UpdateWindowTrigger?.Invoke(window, WindowUpdateType.Move);
+                    WindowUpdated?.Invoke(window, WindowUpdateType.Move);
                 }
             }
         }
 
         private void HandleWindowFocused(IWindow window)
         {
-            WindowFocused?.Invoke(window);
+            OnWindowFocused?.Invoke(window);
         }
 
         private void HandleWindowUpdated(IWindow window)
         {
-            WindowUpdated?.Invoke(window);
+            OnWindowUpdated?.Invoke(window);
         }
 
         private void HandleWindowClosed(IWindow window)
         {
-            WindowClosed?.Invoke(window);
+            OnWindowClosed?.Invoke(window);
         }
 
         private void HandleWindowMoveStart(WindowsWindow window)
@@ -349,12 +349,12 @@ namespace workspacer
 
         private void HandleWindowAdd(IWindow window, bool firstCreate)
         {
-            AddWindowTrigger?.Invoke(window, firstCreate);
+            WindowCreated?.Invoke(window, firstCreate);
         }
 
         private void HandleWindowRemove(IWindow window)
         {
-            RemoveWindowTrigger?.Invoke(window);
+            WindowDestroyed?.Invoke(window);
         }
 
         private bool ShouldIgnoreWindow(IWindow window)
--- a/src/workspacer.Native/Native/WindowsManager.cs
+++ b/src/workspacer.Native/Native/WindowsManager.cs
@@ -30,12 +30,30 @@ namespace workspacer
 
         private Dictionary<WindowsWindow, bool> _floating;
 
+        /// <summary>
+        /// Notifies when a new window handle was created by the manager
+        /// </summary>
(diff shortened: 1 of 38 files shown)


src/workspacer.Native/Native/WindowsWindow.cs
```diff
@@ -163,7 +163,7 @@ public void Focus()
             {
                 Logger.Debug("[{0}] :: Focus", this);
                 Win32Helper.ForceForegroundWindow(_handle);
-                WindowFocused?.Invoke();
+                Focused?.Invoke();
```
josteink, line 166: These changes here looks mostly like change in form, not in function. And they also look like they might break compatibility. Do we really need these changes?

src/workspacer.Shared/Workspace/IWorkspace.cs
```diff
@@ -1,147 +1,127 @@
...
 
         /// <summary>
-        /// the last focused window in the workspace
+        /// The last focused window in the workspace
         /// </summary>
         IWindow LastFocusedWindow { get; }
 
         /// <summary>
-        /// whether the workspace is currently indicating (flashing)
+        /// Whether the workspace is currently indicating (flashing)
```
josteink, line 43: Lots of changes in code-docs makes it harder to identify the actual bits of code changed, and how they have been changed. It would be easier to review this draft without such changes.

src/workspacer.Shared/Workspace/IWorkspace.cs
```diff
@@ -1,147 +1,127 @@
...
-
-        /// <summary>
-        ///  reset the active layout
-        /// </summary>
-        void ResetLayout(); // mod-n
-
-        /// <summary>
-        /// focus the last focused window
-        /// </summary>
-        void FocusLastFocusedWindow();
```
josteink, line 78: Were these methods actually removed, or just alphabetically re-arranged (with code-docs added for some)?

If no actual change was done here, having these changes in among the rest makes reviewing the actual changes harder.

josteink:
Hey there and thanks for the draft.

I'm not that active as a maintainer, but I can see how this kind of functionality can be useful in more complex/orchestrated setups.

As for the code-changes themselves... I really can't comment too much on that. Right now you have lots of changes in a single commit, and it's not immediately obvious which changes are required and which are just "small fixups" you did along the way.

If you leave out fixups and typos and casing-changes, etc, and can reduce this PR to the actual code you want changed to add those new events, I think it will be easier for the other maintainers (which may have a stronger opinion on the underlying changes) to review.

denjukebox:
I've been busy for a bit on other stuff but have come back to this now. I reverted all changes from before and took a new approach without all the unneeded changes.

To address the issue I've expanded the events present in the window class to also provide Update events. This way changes made by other entities can trigger an updated event on the window without having to also include all the manipulation code.

I did have to change some existing naming of events as it was conflicting with the new ones. I hope i made the naming more clear and specific as a result.

src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -14,7 +14,7 @@
 namespace workspacer
 {
     public delegate void WindowDelegate(IWindow window);
-    public delegate void WindowCreateDelegate(IWindow window, bool firstCreate);
+    public delegate void WindowAddDelegate(IWindow window, bool firstCreate);
```
josteink, line 17: Changing the name of this delegate changes a published API. Do we need to do this?

src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -30,10 +30,13 @@ public class WindowsManager : IWindowsManager
 
         private Dictionary<WindowsWindow, bool> _floating;
 
-        public event WindowCreateDelegate WindowCreated;
-        public event WindowDelegate WindowDestroyed;
-        public event WindowUpdateDelegate WindowUpdated;
+        public event WindowAddDelegate AddWindowTrigger;
+        public event WindowDelegate RemoveWindowTrigger;
+        public event WindowUpdateDelegate UpdateWindowTrigger;
```
josteink, line 35: I don't believe it's .NET convention to suffix events with `Trigger`. The conventional form is either:

* OnSomethingHappened
* SomethingHappened

This change breaks .NET conventions *and* changes a published API which may be relied upon by existing configurations out there...

Again: Do we need this change? I'd rather avoid changing things for no reason :smile:

src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -30,10 +30,13 @@ public class WindowsManager : IWindowsManager
...
+
         public event WindowFocusDelegate WindowFocused;
+        public event WindowDelegate WindowUpdated;
+        public event WindowDelegate WindowClosed;
```
josteink, line 39: This may be a stupid question, but technically speaking... What is the difference between `WindowClosed` and `WindowDestroyed`? Might be worth a code-doc?

src/workspacer.Shared/Window/IWindow.cs
```diff
@@ -1,15 +1,18 @@
...
-using System.Diagnostics;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace workspacer
 {
+    public delegate void WindowClosedDelegate();
+    public delegate void WindowUpdatedDelegate();
+    public delegate void WindowFocusedDelegate();
```
josteink, line 8: I believe it's customary/conventional for events like this to have a first argument `object sender` (or in this case `IWindows sender`), to make it easier for the consumer of the event to use and correlate the event-data internally.

src/workspacer.Shared/Workspace/IWorkspaceManager.cs
```diff
@@ -8,10 +8,10 @@ namespace workspacer
...
     public delegate void WorkspaceUpdatedDelegate();
     public delegate void FocusedMonitorUpdatedDelegate();
-    public delegate void WindowAddedDelegate(IWindow window, IWorkspace workspace);
-    public delegate void WindowUpdatedDelegate(IWindow window, IWorkspace workspace);
-    public delegate void WindowRemovedDelegate(IWindow window, IWorkspace workspace);
-    public delegate void WindowMovedDelegate(IWindow window, IWorkspace oldWorkspace, IWorkspace newWorkspace);
+    public delegate void WorkspaceWindowAddedDelegate(IWindow window, IWorkspace workspace);
+    public delegate void WorkspaceWindowUpdatedDelegate(IWindow window, IWorkspace workspace);
+    public delegate void WorkspaceWindowRemovedDelegate(IWindow window, IWorkspace workspace);
+    public delegate void WorkspaceWindowMovedDelegate(IWindow window, IWorkspace oldWorkspace, IWorkspace newWorkspace);
```
josteink, line 14: I don't see the need to rename these delegates (or other types) and break existing configurations out there, unless there are valid reasons for doing so. Are changes required?

src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -298,16 +304,26 @@ private void WindowMove(IntPtr handle)
...
+        {
+            WindowUpdated?.Invoke(window);
+        }
+
+        private void HandleWindowClosed(IWindow window)
+        {
+            WindowClosed?.Invoke(window);
+        }
+
         private void HandleWindowMoveStart(WindowsWindow window)
```
josteink, line 327: Any reason `window` here needs to be a `WindowsWindow` instead of just a `IWindow`?

This particular method stands out :smile:

josteink:
I've gone over the code, and at its core it looks like a useful and welcome change. :heart: 

That said, it still contains changes I don't think is necessary, in particular renamed types, properties and events. This may cause existing users out there to have their configurations broken, and I don't particularly appreciate that.

If you can revert those renames, and just keep the core changes required for this improved functionality, I will be happy to merge this!

src/workspacer.Native/Native/WindowsManager.cs
denjukebox: The issue is that IsMouseMoving is only a Getter in the IWindow class and it is set a few times throughout this class

src/workspacer.Native/Native/WindowsManager.cs
denjukebox: It's only needed not to conflict between eventnames but I see that correctly naming the new events is a better option

src/workspacer.Native/Native/WindowsManager.cs
denjukebox: Code doc for all of the new window events will be added in next day or 2 to clarify the use and origin

denjukebox:
The comments are very on point. I'm still learning the do's and dont's of working on a open source project. That being said I also rename to much in my closed source job so it's definitely a work in progress issue of mine. 

As mentioned I will provide the code documentation in the next few days to clarify the new events

src/workspacer.Native/Native/WindowsManager.cs
josteink: Right. So it's deliberate and not an oversight. Sounds good to me :smile:

src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -33,7 +33,10 @@ public class WindowsManager : IWindowsManager
         public event WindowCreateDelegate WindowCreated;
         public event WindowDelegate WindowDestroyed;
         public event WindowUpdateDelegate WindowUpdated;
-        public event WindowFocusDelegate WindowFocused;
+
+        public event WindowFocusDelegate OnWindowFocused;
```
josteink, line 37: This event is still renamed :)

josteink:
I still found 1 rename, but apart from that this is much clearer and is already looking much, much better.

As for being new to open-source contributions, just take the time you need and don't rush it. We'll get there :smile:

src/workspacer.Native/Native/WindowsManager.cs
denjukebox: Oh woops. I missed that this was a pre-existing one when fixing the "trigger" event naming

src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -30,10 +30,31 @@ public class WindowsManager : IWindowsManager
...
         public event WindowUpdateDelegate WindowUpdated;
+
+        /// <summary>
+        /// Notifies when a window focuses itself
+        /// </summary>
         public event WindowFocusDelegate WindowFocused;
+        /// <summary>
+        /// Notifies when a window updated itself
+        /// </summary>
+        public event WindowDelegate OnWindowUpdated;
```
josteink, line 53: In what ways is this event different enough from the existing `WindowUpdated` event that it needs to be special cased?

Just curious! :smile:

josteink: Right. Reading the code, it seems we have this event specifically to be able to further trigger manager-based `WindowUpdated` events without causing recursion and event-looping?

If so, that's fine by itself. I'm just wondering if there would be a way to structure the code to allow this without having two seemingly similar events like this. Any opinions?

denjukebox: As you indeed correctly, using the existing event in its current state causes a feedback loop of updates (found that out the hard way). As the logic path for this event is very specific to processing updates on a window within a workspace I think merging the two would cause more confusion, though the suggestion would be adding another bool param to the event like "doWindowHandling" with default true which would skip the handling code when false.

josteink: If we either make it clear that one event is for *internal* changes or that the other event is for *externally* triggered changes.... That would be a good enough distinction for me.

Since one event already was there from the get go, I think it makes sense to tag this new event as being the "external" one.

It's a bit verbose and rich... But how about `WorkspacerExternalWindowsUpdate`? It's only registered one place after all?

denjukebox: I liked the suggested fix so i applied it!

josteink (approved):

josteink:
Thanks for hanging in there for so long. This PR looks good to me now, so consider it merged. :smile: 

Thanks for contributing!

PR #276, opened 2021-12-01
Feat/focus layout

blalande:
Add a focus layout, with a central primary area and 2 sides secondary areas

### DIFF
--- a/src/workspacer.Shared/Layout/FocusLayoutEngine.cs
+++ b/src/workspacer.Shared/Layout/FocusLayoutEngine.cs
@@ -1,0 +1,115 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace workspacer
+{
+    // a layout that put the master view in the center of the screen, and the secondary windows on both sides, new windows opening on the right side
+    public class FocusLayoutEngine : ILayoutEngine
+    {
+        public string Name => "focus";
+
+        private readonly int _numInPrimary;
+        private readonly double _primaryPercent;
+        private readonly double _primaryPercentIncrement;
+
+        private int _numInPrimaryOffset = 0;
+        private double _primaryPercentOffset = 0;
+
+
+        public FocusLayoutEngine() : this(1, 0.7, 0.03) { }
+
+        public FocusLayoutEngine(int numInPrimary, double primaryPercent, double primaryPercentIncrement)
+        {
+            _numInPrimary = numInPrimary;
+            _primaryPercent = primaryPercent;
+            _primaryPercentIncrement = primaryPercentIncrement;
+
+        }
+
+        public IEnumerable<IWindowLocation> CalcLayout(IEnumerable<IWindow> windows, int spaceWidth, int spaceHeight)
+        {
+            var list = new List<IWindowLocation>();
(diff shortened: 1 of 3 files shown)


josteink:

I haven't tested this, but the code itself looks good!

On my part I have no obvious objections, especially considering that this is a layout-engine and that the code in it is 100% self-contained and doesn't affect any other modules we have.

I see there's some math-work being done inside the `CalcLayout()` method... Not sure if that could/should be extracted to be more unit-testable (and thus prove that the engine works as expected)? On the flip side the calculations depends on quite a few variables, it extracting the calculation into smaller functions might make the code harder to read?

I don't know. What do you tink? :smile:

blalande:
Hi,
I agree, the calculation can probably be isolated into function. I’ll try to clean that up.
Thanks for reviewing ! For my first PR I tried something simple, I have others ideas that may impact more stuff... I’ll be happy to discuss them :)

src/workspacer.Shared/Layout/FocusLayoutEngine.cs
```diff
@@ -78,6 +78,16 @@ public IEnumerable<IWindowLocation> CalcLayout(IEnumerable<IWindow> windows, int
...
 
+        private int GetLeftHeight(int spaceHeight, int numWindows, int numInPrimary)
+        {
+            return spaceHeight / Math.Max((numWindows - numInPrimary) / 2, 1);
+        }
+
+        private int GetRightHeight(int spaceHeight, int numWindows, int numInPrimary)
+        {
+            return spaceHeight / Math.Max((numWindows - numInPrimary + 1) / 2, 1);
+        }
```
josteink, line 89: Looking at this code here, I'm trying to do some mental math to make sure I understand the formulas used.

Let's say we have:

* numWindows = 5
* numInPrimary = 2
* spaceHeight = 800

That gives us:

* windowsOutsidePrimary = 5 - 2 = 3
* numLeftWindows = Max((windowsOutsidePrimary/2), 1) = 1.5, except int, so no rounding = 1
* numRightWindows = Max(((windowsOutsidePrimary +1 )/2), 1) = 2

Which leaves us numLeftWindows + numRightWindows = 3. Which adds up to windowsOutsidePrimary. Great!

Now lets try with:

* numWindows = 6
* windowsOutsidePrimary = 6 - 2 = 4
* numLeftWindows = Max((windowsOutsidePrimary/2), 1) = 2
* numRightWindows = Max(((windowsOutsidePrimary +1 )/2), 1) = 2.5 except int, so no rounding = 2

And it still adds up. Right!

But I had to do all that math myself, and make up those test-cases, because there's no unit-tests for this bit right now.

Wouldn't it be nice to have tests for that, so that we know it works now and will keep working in the future? :)

blalande: Thanks for all those suggestions. I’ve isolated the calculation of the number of windows for the 2 secondary panel. I’ve also added unit testing for those. And fixed some bugs and weird behavior

josteink (approved):

josteink:
Looks good to me. Consider it merged! Thanks for hanging in there :smile:

PR #295, opened 2022-03-05
Added function to print logging info for active window

derekthecool:
Workspacer is already keyboard oriented, this should be a nice addition
for anyone who wants to use just keyboard.

I found myself wanting this feature while setting up routes, and getting annoyed by using the mouse.
Also getting annoyed by having to scroll up through the entire list of windows.

### DIFF
--- a/src/workspacer.Native/Native/WindowsManager.cs
+++ b/src/workspacer.Native/Native/WindowsManager.cs
@@ -91,15 +91,14 @@ namespace workspacer
 
         public void DumpWindowDebugOutputForFocusedWindow()
         {
-            var output = "";
             var focusedWindow = Windows.Where(win => win.IsFocused)
                 .Select(win => win).FirstOrDefault();
 
             if (focusedWindow != null)
             {
-                output += GenerateWindowDebugOutput(focusedWindow) + "\r\n\r\n";
+                var output = GenerateWindowDebugOutput(focusedWindow);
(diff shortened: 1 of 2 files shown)


src/workspacer.Native/Native/WindowsManager.cs
```diff
@@ -89,6 +89,19 @@ public void DumpWindowDebugOutput()
...
 
+        public void DumpWindowDebugOutputForFocusedWindow()
+        {
+            var output = "";
+            var focusedWindow = Windows.Where(win => win.IsFocused)
+                .Select(win => win).FirstOrDefault();
+
+            if (focusedWindow != null)
+            {
+                output += GenerateWindowDebugOutput(focusedWindow) + "\r\n\r\n";
```
josteink, line 100: Output has previously only been assigned a blank value, and we don't need to concatenate to it. `output = ` alone will be fine.

Also: Any reason we need to append those newlines at the end? :smile:

josteink:
Hey there and thanks for the patch!

Changes looks mostly good to me, but I see there are a few small things which can be cleaned up before we merge :)

src/workspacer.Native/Native/WindowsManager.cs
derekthecool: Done

josteink (approved):
LGTM!

PR #299, opened 2022-04-04
chore: rearrange tray menu item

amrbashir:
It is commonly expected that the `Quit` menu item is the last one in the list and `Restart` should probably be directly above `Quit`

### DIFF
--- a/src/workspacer/Workspacer.cs
+++ b/src/workspacer/Workspacer.cs
@@ -39,10 +39,10 @@ namespace workspacer
             });
 
             // init system tray
-            _context.SystemTray.AddToContextMenu("enable/disable workspacer", () => _context.Enabled = !_context.Enabled);
-            _context.SystemTray.AddToContextMenu("quit workspacer", () => _context.Quit());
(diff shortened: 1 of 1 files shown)


josteink:
Hey there and sorry for not looking into this PR.

The changes looks good to me. If you are willing to re-open, I'll get this merged :smile:

amrbashir:
No need to apologize, It is my bad, I thought the project wasn't actively maintained anymore. The PR had a mistake anyway, I will recreate it.

josteink:
It's not *very* maintained to be honest. I'm honestly not sure where the other maintainers are. We used to have 3-4 in total.

I'm on Linux these days (using swaywm), so I don't use Workspacer anymore. That said, it seems to work for quite a few people, so I try to keep it somewhat alive by helping out where I can.

So I pop in every now and then to review simple PRs, but since I'm not running Workspacer myself, I only approve PRs I can assert are correct based on code review alone, without having to manually run and validate them ;)

That does limit the scope of changes being introduced, but at least it's not complete dead :smile:

PR #309, opened 2022-05-10
Fix #308 - docs build error

### LINKED ISSUE #308: Creating docs fails - breaks the build and site-deploy
See any recently merged PR, and verify the build:

https://github.com/workspacer/workspacer/actions/runs/2299941789

You'll see they all fail at the same step, running `hugo` to update the docs site:
image

I'm not really that deep into Hugo. Are there anyone who can look into this? :smile: 

CC: @BeeryShklar @dalyIsaac

BeeryShklar:
This should fix #308.
As mentioned [here](https://discourse.gohugo.io/t/pages-url-is-deprecated/20645), the Page.URL variable in Hugo was deprecated and got removed so I changed everywhere it appeared to `.Params.URL`.

### DIFF
--- a/docs/themes/workspacer/layouts/partials/footer.html
+++ b/docs/themes/workspacer/layouts/partials/footer.html
@@ -2,7 +2,7 @@
 	<p>&copy; Licence MIT.</p>
 	<nav class="nav-footer">
 		{{ range .Site.Menus.footer }}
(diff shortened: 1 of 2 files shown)


josteink:
Great. Thanks for helping out! Consider it merged :)

josteink (approved):

PR #324, opened 2022-06-25
New unstable is 0.9.12 now.

josteink:
I made a 0.9.11 stable release in response to the Newtonsoft.Jon vulnerability (old stable had vulnerable version).

As such, we need to bump the version for the unstable version :smile:


josteink:
Nobody around to review these days. Merging myself :)

PR #328, opened 2022-07-26
Add explicit reference to update Netwtonsoft.Json in all test-projects

josteink:
Fix warnings about insecure version of Newtonsoft.Json.


josteink:
If it builds, it builds. Merging now!

PR #329, opened 2022-07-29
Plugin: Bar sound widget and control

denjukebox:
When using workspacer as replacement for the windows shell, the sound key binds (volume/mute) are lost. This PR adds back this functionality using the following library [NetCoreAudio](https://github.com/dorba/netcoreaudio) that provides all the hooks into windows. It uses the same MIT licence as workspacer so usage is permitted. Beside this it also provides a bar widget that can follow default device changes and is configurable and expandable it its presentation.


josteink:
This is a big serious PR! Nice one!

I really don't think I can do a full test of the functionality here, given how I no longer run Windows as my primary OS. That said, this is too much good work to pass off :smile: 

So I'll try to come up with some constructive feedback, and if no other maintainer shows up, we'll just merge it none the same.

*If* this change for whatever reasons causes isseus, whoever runs "unstable" channel should be prepared for things not being 100% reliable anyway :smile:

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+
+            DevicePropertyValueChanged?.Invoke(new DeviceInfo { Id = deviceId }, propertyKey);
+#if EVENT_DEBUG
+            MessageBox.Show($"DeviceStateChanged deviceId:[{deviceId}] propertyKey:[{propertyKey}]");
+#endif
+        }
+        #endregion
+    }
+
+    public class SoundClient : IAudioEndpointVolumeCallback
```
josteink, line 213: I see the `SoundClient` class being exposed to the "outside world" through the `SoundManager` class.

As such, I think it would be nice to put in its own file. As a general rule, I think that's always nice.

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+
+            return client;
+        }
+
+        public SoundClient RegisterClientForDeviceId(string deviceId)
+        {
+            return RegisterClient(deviceId);
+        }
+
+        private SoundClient RegisterClient(string deviceId, DeviceType? type = null, params DeviceRole[] roles)
```
josteink, line 52: Semantically `RegisterClient(...)` very much sounds like you as a caller are expected to provide a client of some sort... And it also sounds like something you may have to "unregister" later. But from what I see, there's no "unregister" API, so currently it feels like it's somewhat lacking a symmetry of sorts.

But given that we `RegisterClient(...)` don't expect a client at all, but rather *returns a "client" to the caller*, I think a better naming-convention for these methods would be `GetClientForXXX(...)`. How about it?

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+        {
+            var client = CreateClient(deviceId, type, roles);
+
+            //Map events
+            client.VolumeChanged += HandleVolumeChanged;
+            client.MuteChanged += HandleMuteChanged;
```
josteink, line 58: This is the only place where `CreateClient(...)` is called, and it seems to mostly prepare the object for use. But then immediately afterwards you start configuring the returned object further.

I think moving the registration of these event-handlers into `CreateClient()` would make the code cleaner.

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+        }
+
+        public string GetDefaultRecordingDeviceId(DeviceRole role = DeviceRole.Communications)
+        {
+            return GetDefaultDeviceId(DeviceType.Capture, role);
+        }
+
+        public string GetDefaultDeviceId(DeviceType type, params DeviceRole[] roles)
+        {
+            if (_mmDeviceEnumerator == null)
```
josteink, line 92: Can this condition ever occur in a working instance? If not, then we don't need this check.

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+                return null;
+            }
+
+            IMMDevice device = null;
+            foreach (var role in roles)
+            {
+                _mmDeviceEnumerator.GetDefaultAudioEndpoint(type.ConvertToEDataFlow(), role.ConvertToERole(), out device);
+                if(device != null)
+                {
+                    break;
```
josteink, line 103: It may be a matter of taste, but simply retuning the device-id inside this if-block would improve code locality, reduce the amount of intermediate variables created, and make it more obvious what happens in the 2 different use-cases (device found, device not found).

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+            }
+
+            var id = string.Empty;
+            device?.GetId(out id);
+            return id;
+        }
+
+        private SoundClient CreateClient(string deviceId, DeviceType? type = null, params DeviceRole[] roles)
+        {
+            if (_mmDeviceEnumerator == null)
```
josteink, line 114: Again. Do we need this check?

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+            {
+                return null;
+            }
+
+            _mmDeviceEnumerator.GetDevice(deviceId, out var device);
+            if (device == null)
+            {
+                return null;
```
josteink, line 122: Is this expected to happen under normal operations? That is, for an incorrect device-id to be passed in? And if so, are we sure simply returning `null` is the most appropriate way to handle it?

Not saying we *must* change this, but I'm just curious about what sort of considerations was put into this use-case, since it was so clearly handled :smile:

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+
+        private IAudioEndpointVolume _deviceAudioEndpointVolume;
+
+        public void SetDevice(IMMDevice device)
+        {
+            _device = device;
+            InitializeAudioEnpoint(device);
+        }
+
+        private void InitializeAudioEnpoint(IMMDevice device = null)
```
josteink, line 230: Do we need this argument to be optional? Do we need it at all? When would be initialize another audio-endpoint than the one the object represents?

src/workspacer.Sound/SoundManager.cs
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }
+
+        //Notify on default device change
+        public int OnNotify(IntPtr notificationData)
+        {
+
+#if EVENT_DEBUG
+            MessageBox.Show($"OnNotify {notificationData}");
+#endif
```
josteink, line 250: These ifdefs is probably not going to make it into git master, right? :smile:

workspacer.sln
```diff
@@ -35,6 +35,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "workspacer.TitleBar", "src\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "workspacer.TitleBar.Tests", "src\workspacer.TitleBar.Tests\workspacer.TitleBar.Tests.csproj", "{E4E357BA-0154-4266-A816-B765208E810D}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "workspacer.Sound", "src\workspacer.Sound\workspacer.Sound.csproj", "{80D9E2E0-5A4F-4DD1-AB21-7535DF162A75}"
+EndProject
```
josteink, line 39: You've added a new project. Great!

Do you think any of the new classes in this project would benefit from tests to ensure the adhere to defined behaviour?

I hate seeing new code added without *any* tests :crying_cat_face:

src/workspacer.Sound/SoundPluginConfig.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace workspacer.Sound
+{
+    public class SoundPluginConfig
+    {
+        public bool BindDefaultPlaybackKeybinds { get; set; }
```
josteink, line 5: In this case, default = `false`, which should result in the behaviour you are complaining about being fixed... But then it can be overridden if the user wants to. Right? If so, that's good.

src/workspacer.Sound/workspacer.Sound.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<ItemGroup>
+		<ProjectReference Include="..\workspacer.Bar\workspacer.Bar.csproj" />
+		<ProjectReference Include="..\workspacer.Shared\workspacer.Shared.csproj" />
+	</ItemGroup>
+	<ItemGroup>
+	  <Reference Include="Vannatech.CoreAudio">
+	    <HintPath>Vannatech.CoreAudio.dll</HintPath>
```
josteink, line 8: I'm not sure I like adding binaries to the project like this.

I would be much more comfortable with this if it was provided as a nuget-packages.

Can't you nudge the maintainers over at https://github.com/dorba/netcoreaudio to release this as a package?

src/workspacer/workspacer.csproj
```diff
@@ -27,7 +27,8 @@
     <ProjectReference Include="..\workspacer.Gap\workspacer.Gap.csproj" ReferenceOutputAssembly="false" />
     <ProjectReference Include="..\workspacer.ActionMenu\workspacer.ActionMenu.csproj" ReferenceOutputAssembly="false" />
     <ProjectReference Include="..\workspacer.FocusIndicator\workspacer.FocusIndicator.csproj" ReferenceOutputAssembly="false" />
-    <ProjectReference Include="..\workspacer.TitleBar\workspacer.TitleBar.csproj" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\workspacer.Sound\workspacer.Sound.csproj" />
```
josteink, line 30: Just checking if this is deliberate or not... In this case we *want* to reference the output assembly, right?

src/workspacer.Sound/workspacer.Sound.csproj
denjukebox: Definitely agree, would be a good addition

src/workspacer.Sound/SoundPluginConfig.cs
denjukebox: Well I suspect most workspaces users would use this only for the widget and not the actual control and bindings. That's why I left this default false so no double actions would happen on the media sound controls

workspacer.sln
denjukebox: Yes definitely, I'll add tests for the situations now that I know this functionality is wanted in main workspacer

src/workspacer.Sound/SoundManager.cs
denjukebox: Will remove! To be honest I forgot about these after testing

marad:
> When using workspacer as replacement for the windows shell

@denjukebox You mean, like... no windows shell at all? How to do that?

src/workspacer.Sound/workspacer.Sound.csproj
josteink: That repo seems positively abondoned. No new activity in roughly 4 years. This fork seems more maintained though?

https://github.com/vuplea/CoreAudioAPI

josteink: It's not very maintained, and doesnt have a nuget-package published.

How about using this library which seems to be popular (and maintained)?

https://www.nuget.org/packages/NAudio

jbrazio:
This was common practice on the Windows NT/XP world, it was just a question of replacing a registry key. I've searched this topic for Windows 11 and it seems a bit more complicated but still doable.

https://learn.microsoft.com/en-us/windows/configuration/kiosk-shelllauncher

PR #339, opened 2022-08-25
Add href references to links in header.

tlsomers:


josteink:
Nice catch. Consider it merged!

Thanks for contributing!

josteink (approved):

BeeryShklar:
This is only gonna fix the issue temporarily. Next time the site is compiled the changes will be overridden.

I'm pretty sure this was caused by the last PR I made (#309) so I will try and permanently fix this as soon as I can.

PR #340, opened 2022-08-25
Permanently fix href references in links

BeeryShklar:
#339 only fixed the bug in the output files which are gonna be overridden when a new version of the site is created. This should fix the issue permanently.
Thanks @tlsomers for catching this. Nice work!


josteink (approved):

josteink:
That was quick. Thanks for helping out!

PR #349, opened 2022-09-12
Create plugin for allowing configuration in F#

tlsomers:
Adds a plugin to allow the extending of a C# configuration with an F# configuration.
Compiles the F# script to a DLL after changes to provide faster loading (0.1s to load DLL vs 6s to compile or interpret)

Future work:
- Move C# configuration out of main project
- Configure which plugin to use for configuration (for instance using json or use first existing)

Benefit:
- F# scripts are older and provide autocompletion in both Visual Studio and Rider (unlike C# scripts)


josteink:
This PR in its current state contains lots of code which is mixed in from another PR.

Please create a clean PR (or update this one) with only the F# stuff, so that can be reviewed in isolation.

PR #222, opened 2021-06-29
Additional configuration options for workspacer.Bar

N1x0:
This PR aims to introduce more configuration options for the `BarPlugin`, to allow changing the Bar's width and position as a fix to #212 
Adds the following features:

* `bool BarIsTop` sets the position of the bar
  * Added in:  102139f2bf6bc9ef8a63f98ea3cf823c846851ea
  * Default is `true`

* `bool BarIsMinimal` sets the bar width to the width of all widgets 
  * Added in:  468751354694da71eb895198dad0d64f833e7730
  * Default is `false`

* `bool BarReservesSpace` sets the offset in the `MenubarLayoutEngine`
  * Added in: bb8db8575fa33163dbb8d89686dc8aa760cae568
  * Default is `true`

Tasks
 - [ ] dynamically adjust barwidth at runtime to avoid painting over other widgets.
 - [ ]  Force the bar  to stay on top when no space is being reserved for it
 - [ ]  Move the bar in bottom position when the windows taskbar is auto-hiding

PR #353, opened 2022-09-12
Additional configuration options for workspacer.Bar

tlsomers:
Because of the inactivity of Pull Request https://github.com/workspacer/workspacer/pull/222, I will be attempting to pick up and continue with the feature. These are my first fixes.


src/workspacer.Bar/BarPluginConfig.cs
```diff
@@ -19,6 +20,11 @@ public class BarPluginConfig
 
         public Color DefaultWidgetForeground { get; set; } = Color.White;
         public Color DefaultWidgetBackground { get; set; } = Color.Black;
+        
+        public Color Transparant { get; set; } = Color.Lime;
```
josteink, line 24: I don't think the name of this property is enough to convey it's meaning. Wouldn't using the same name as elsewhere (i.e. `TransparencyKey`) be more obvious?

tlsomers: You're right, I updated it accordingly

PR #356, opened 2022-09-16
Add support for configuration in F#.

tlsomers:
This pull request replaces: https://github.com/workspacer/workspacer/pull/349

Future work: Actually use the IConfig interface and possibly move C# configuration out of main project.


josteink:
Thanks for cleaning this PR up.

As a high-level comment (without looking into the code itself), this is a lot of code, for a possibly new feature... But there's no documentation for it.

Would it be possible to add a section to the quickstart with a short example of/note about how to use this feature?

That would also make it easier (for a non-regular F#ie) to put the code in the PR into context.

tlsomers:
I'll add a section to the documentation as soon as I find the time, until then this pr can be left as is, I'll let you know when the documentation is done.

tlsomers:
@josteink I have now updated the quickstart with some information on using the F# configuration feature.

docs/content/quickstart.md
```diff
@@ -48,6 +48,35 @@ Workspacer's configuration file allows for __type checking__ and __intellisense_
...
+```
+C:\Users\<username>\.config\workspacer\workspacer.config.csx
+```
+
+With this configuration, workspacer is configured in the `setupContext` function in the F# configuration script:
+```
+C:\Users\<username>\.config\workspacer\Workspacer.Config.fsx
+```
+
+An example configuration, equivalent to the default C# configuration is available on [github](TOOD). In the F# configuration, it is important to use `#r @".."` directives instead of `#r ".."`. Furthermore, a function `setupContext` is used instead of the returned action in the C# configuration. These differences are due to the F# configuration being compiled to a `dll` library, rather than running as a script directly.
```
josteink, line 76: TODO in link needs to be fixed? :)

src/workspacer.FSharpConfig/Extensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Reflection;
+
+    public static class Extensions
+    {
+        public static Assembly LoadFile(this AppDomain appDomain, string filename) {
+            FileStream fs = new FileStream(filename, FileMode.Open);
```
josteink, line 12: Using this FileStream constructor may cause less conflicts with file being used or not :smile: 

https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.-ctor?view=net-7.0#system-io-filestream-ctor(system-string-system-io-filemode-system-io-fileaccess-system-io-fileshare)

src/workspacer.FSharpConfig/Extensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            byte[] buffer = new byte[(int) fs.Length];
+            fs.Read(buffer, 0, buffer.Length);
+            fs.Close();
```
josteink, line 15: Any reason not to just use `var bytes = File.ReadAllBytes(filename)` ?

josteink: Reading a stream like this is known to be unreliable, and you should always return the *actual* amount of bytes read returned from the `FileStream.Read()`-invocation.

src/workspacer.FSharpConfig/Extensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            return appDomain.Load(buffer);
+        }
+
+        public static AppDomain FileResolver(this AppDomain resoleAppDomain, IEnumerable<string> dllFiles)
```
josteink, line 20: This is a method and should have a verb-y name, but FileResolver is a noun. If it's function is to idempotently load dll-files... Then why not simply call it `LoadFiles()`?

Also the extended AppDomain instance has a typo in it (resolveAppDomain), Since this is an extension method, I think a more appropriate name would simply be `instance`.

src/workspacer.FSharpConfig/Extensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            var references = dllFiles.ToDictionary(x => AssemblyName.GetAssemblyName(x).FullName);
+
+            Assembly Resolver(object sender, ResolveEventArgs args)
+            {
+                AppDomain domain = (AppDomain) sender!;
+
+                var file = references[args.Name];
+                return LoadFile(domain, file);
+            }
```
josteink, line 30: I think declaring this event-handler function here instead of directly attaching it to the event makes it much more confusing to read.

src/workspacer.FSharpConfig/Extensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public static T GetDelegate<T>(this Type t, string name)
+            where T : Delegate
+        {
+            var method = t.GetMethod(name);
+            return (T)Delegate.CreateDelegate(typeof(T), method);
+        }
+
+        public static T GetDelegate<T>(this Assembly a, string t, string name)
```
josteink, line 43: I'm not sure how this function is used, but if Type is known at compile-time... Would it make more sense to accept this as a second type-parameter?

````c#
public static TDelegate GetDelegate<TType, TDelegate>(this Assembly  a, string name)
````

?

File: src/workspacer.FSharpConfig/FSharpCompileException.cs [hunk 7]
```diff
@@ -0,0 +1,15 @@
+﻿namespace workspacer.FSharpConfig
+{
+    using System;
+    using FSharp.Compiler.Diagnostics;
+
+    public class FSharpCompileException : Exception
+    {
+        public FSharpDiagnostic[] Errors { get; }
+
+        public FSharpCompileException(string message, FSharpDiagnostic[] diagnostics) : base(message)
```
josteink, line 10: If the property is called `Errors`, it would be more consistent to name the parameter `errors` as well.

src/workspacer.FSharpConfig/FSharpCompiler.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                compilerArgs.Add($"--lib:\"{r}\"");
+            }
+
+            var result =
+                FSharpAsync.RunSynchronously(
+                    checker.Compile(compilerArgs.ToArray(), FSharpOption<string>.None),
+                    FSharpOption<int>.None,
+                    FSharpOption<CancellationToken>.None
+                );
```
josteink, line 52: As far as I can see, this compiles and runs the config in one go.

Would it make more sense to run the compiler first, and *then* run the code in a subsequent call?

That might make it easier to handle and present meaningfull diagnostics-messages when the configuration either contains compilation-errors or has runtime-failures.

src/workspacer.FSharpConfig/FSharpCompiler.cs
```diff
@@ -0,0 +1,94 @@
...
+            if (result.Item2 != 0)
+            {
+                throw new FSharpCompileException(result.Item1[0].Message, result.Item1);
```
josteink, line 55: Here we seem to imply all errors when running the F# config are compilation errors.

How do we deal with runtime errors? Are they handled automagically via .NET interop?

src/workspacer.Shared/IConfigContext.cs
```diff
@@ -23,6 +23,8 @@ public interface IConfigContext
 
         IMonitorContainer MonitorContainer { get; set; }
         bool CanMinimizeWindows { get; set; }
+        
+        string UserWorkspacerPath { get; }
```
josteink, line 27: Any reason we need this new property here?

src/workspacer/Workspacer.cs
```diff
@@ -86,6 +89,7 @@ public void Start()
             _context.Plugins.AfterConfig(_context);
 
             // start message pump on main thread
+            Logger.Info($"Elapsed {sw.ElapsedMilliseconds}ms");
```
josteink, line 92: We may want to log WHAT elpased for so many milliseconds :)

src/workspacer.FSharpConfig/Extensions.cs
tlsomers: In the way this is used the type is not known at compile time, just the name of the type. This is the class the F# script is compiled into (the same name as the file).

src/workspacer.Shared/IConfigContext.cs
tlsomers: Yes, this property is required to retrieve the F# configuration file, as the location is determined in the 'workspacer' project and therefore cannot be used directly.

src/workspacer.FSharpConfig/Extensions.cs
josteink: Fair enough.

src/workspacer.FSharpConfig/FSharpCompiler.cs
josteink: Then I guess this OK as is.

src/workspacer.Shared/IConfigContext.cs
josteink: Right. In which case, I guess my main complaint would be about the naming off this property.

What does `UserWorkspacerPath` imply? I don't this name fully conveys that. Is it for the configuration-folder? For the currently running binaries?

I think we need to make this property more self-explanatory :smile:

tlsomers:
In more testing I discovered a quite bad issue with loading System types such as arrays. Being unsuccessful in fixing it I have put this in draft and on hold.

josteink:
No problem. Let me know when you have something new to review.

src/workspacer.Shared/IConfigContext.cs
josteink: That definitely sounds better to me. I'm all for it.

tlsomers:
The issue is mainly with compiling the wrong system libraries (as you can't choose apparently), so the solution is to resolve those assemblies to the already loaded version.

src/workspacer.FSharpConfig/Extensions.cs
```diff
@@ -0,0 +1,67 @@
+﻿namespace workspacer.FSharpConfig
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Reflection;
+
+    public static class Extensions
```
josteink, line 9: Since this class extends the `AppDomain`-class, I think a better name for it would be `AppDomainExtensions`.

src/workspacer.FSharpConfig/FSharpConfig.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            return Extend;
+        }
+
+        /// <summary>
+        /// Extend a currently executing C# configuration with the F# configuration file. Called from within the
+        /// C# configuration action.
+        /// </summary>
+        /// <param name="context">The configuration context.</param>
+        public static void Extend(this IConfigContext context)
```
josteink, line 39: Does it really make sense to have this as an extension method? Wouldn't a regular static method do?

Especially as this will now pop us as an option for all context-instances, and simply be called `Extend()`, but it will only load the F# config.

For someone not involved in F# at all, this part does at all not seem obvious.

src/workspacer.FSharpConfig/FSharpConfigPlugin.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        /// <summary>
+        /// Does nothing for the moment. In the future we may add an extra menu option to create a template.
+        /// </summary>
+        /// <param name="context">The configured configuration context.</param>
+        public void AfterConfig(IConfigContext context)
+        {
+        }
+    }
+}
```
josteink, line 16: For now this class does nothing, and provides nothing.

If we need something like this in the future, can't we add it then, instead of adding some empty boilerplate now? :smile:

josteink:
Looking pretty good now IMO. That said, I still have a few small nitpicks.

With those sorted out, I think this PR should be ready for merging.

src/workspacer.FSharpConfig/FSharpConfigPlugin.cs
tlsomers: Sadly it's mandatory to have such a class in the current plugin loader 😅

josteink (approved):
LGTM

Janthor:
I'd like to move over to this kind of configuration, mainly for syntax highlighting and other QOL features this brings, but my csx config is pretty complex right now, and I'd like to know if there are plans of providing a more indepth configuration example for people moving over that don't exactly know the direct translations from an C# to an F# environment. Here's [my current config](https://github.com/Janthor/windotfiles/blob/master/workspacer/workspacer.config.csx)

tlsomers:
@Janthor I don't currently have plans for adding extra documentation on translations to F#. I did take the liberty to convert most (I don't have the aliases) of your configuration to F#. Here is a link to the file.

Here is a link to the F# configuration, please change the paths in the `#r @"..."` directives to match your system

Note that conversion to F# is relatively straightforward:
1) convert all `var` and other variable declarations to `let`.
2) convert all mutable assignments to use `<-` instead of `=`,  bitwise or to ||| and logical not to `not`
3) replace all lambda expressions to F# functions (and surround in brackets if necessary) (`x => y` becomes `fun x -> y`)
4) change definitions of arrays to use  [| |] and remove commas in lists (newline is new item)

tlsomers:
image

Also, that's a lovely bar

Janthor:
it's even more lovely if you install a nerd font 👀. Thanks a ton for the explanation, I'd love to have my config in one of the examples too. I'll update it even more over time.

Janthor:
hate to be commenting on a merged PR for this, but can u give me the missing link that didn't come through here @tlsomers ? 
> 
> Here is a link to the F# configuration, please change the paths in the `#r @"..."` directives to match your system
> 
I tried following your brief instructions but I hit a roadblock while declaring the bar, and I feel a practical example would make me understand it better.
This is what I came up with, but I feel i'm doing something extremely wrong lmao
```fsx
        //* Left Widgets
        LeftWidgets <- fun () -> new IBarWidget[||] =
        
            new ActiveLayoutWidget()
            new TextWidget("|")
            new WorkspaceWidget() = 
                WorkspaceHasFocusColor <- positive
                WorkspaceEmptyColor <- alt_background
                WorkspaceIndicatingBackColor <-  urgent
```
Hit me up on Discord if you want to keep the github clean Janthor#3673 (a public Workspacer server on Discord would be great also)

tlsomers:
Here's the relevant gist:
https://gist.github.com/tlsomers/52fc7da2425bfd2caad4393ce3791e7f
I also sent you a request if you need any more help :)

PR #348, opened 2022-09-11
fix OverflowException on Win11 x64

KenMercusLai:
Fix the exception when running on Windows 11 x64:
Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.IntPtr.op_Explicit(IntPtr value)
   at workspacer.ConsoleHelper.get_IsConsoleShowing() in D:\a\workspacer\workspacer\src\workspacer.Watcher\ConsoleHelper.cs:line 35
   at workspacer.ConsoleHelper.set_IsConsoleShowing(Boolean value) in D:\a\workspacer\workspacer\src\workspacer.Watcher\ConsoleHelper.cs:line 46
   at workspacer.ConsoleHelper.Initialize() in D:\a\workspacer\workspacer\src\workspacer.Watcher\ConsoleHelper.cs:line 25
   at workspacer.Watcher.Program.Main(String[] args) in D:\a\workspacer\workspacer\src\workspacer.Watcher\Program.cs:line 19

PR #357, opened 2022-09-16
Resolve Win64 overflow on console toggle.

tlsomers:
Reimplementation of pull request https://github.com/workspacer/workspacer/pull/348.

The setter was fine on the overflow side as it does not make use of IntPtr (which can be both an int pointer and long pointer).
Also fixed the setter which was acting as a 'toggle' rather than using the provided value.


josteink:
LGTM! Thanks for helping out!

PR #359, opened 2022-09-20
No message by Display Settings triggered restart

Janthor:
Removes the message and dialog box for when workspacer gets restarted due to changes in Display Settings (connecting and disconnecting monitors, going into exclusive fullscreen with non-native resolutions, changing display resolution, etc..). 

Current message is redundant since I see no reason to quit workspacer because of those reasons, user usually wants to have workspacer back ASAP, and that dialog delays this process.

### DIFF
--- a/src/workspacer/ConfigContext.cs
+++ b/src/workspacer/ConfigContext.cs
@@ -228,8 +228,7 @@ namespace workspacer
             SaveState();
             var response = new LauncherResponse()
             {
-                Action = LauncherAction.RestartWithMessage,
-                Message = "A display settings change has been detected, which has automatically disabled workspacer. Press 'restart' when ready.",
+                Action = LauncherAction.Restart
             };
             SendResponse(response);
 


Janthor:
Now that I thought more about the matter, maybe following AwesomeWM's  way of doing things may be the way to go... Defaults to restart with message, but exposes this function so the user can change its behavior in their config file, so that they can run whatever they want when the Geometry/Display Settings changes.
this is how it looks in the user config file in awesome:
```lua
-- Re-set wallpaper when a screen's geometry changes (e.g. different resolution)
screen.connect_signal("property::geometry", Set_wallpaper)
```
or
```lua
--Restart awesome when a screen's geometry changes
screen.connect_signal("property::geometry", awesome.restart)
```

tlsomers:
A few uses I can think of for this are:

- Assigning workspaces to monitors after restart based on preference (ex 1 on left monitor, 2 on right)
- changing layout of current workspaces on monitors (ex full screen on main, but tall on secondary)
- Automatically starting/stopping applications on a specific monitor. For instance I always have discord or teams open if I use 3 screens

Janthor:
> A few uses I can think of for this are:
> 
> * Assigning workspaces to monitors after restart based on preference (ex 1 on left monitor, 2 on right)
> * changing layout of current workspaces on monitors (ex full screen on main, but tall on secondary)
> * Automatically starting/stopping applications on a specific monitor. For instance I always have discord or teams open if I use 3 screens

these are actually really good use cases and i'd probably use them lmao, moving to draft an i'll try implementing it

Janthor:
After a lot of fumbling around, I just noticed that as soon as a restart happens (and it needs to happen for workspacer to work), you can handle all of the following things in the top of your config file, since it's going to be ran again. So I guess just not showing the restart message really is enough. Going back to my original idea and taking this out of the draft state.

> A few uses I can think of for this are:
> 
> * Assigning workspaces to monitors after restart based on preference (ex 1 on left monitor, 2 on right)
> * changing layout of current workspaces on monitors (ex full screen on main, but tall on secondary)
> * Automatically starting/stopping applications on a specific monitor. For instance I always have discord or teams open if I use 3 screens

josteink:
I have no strong opinion on this change. 

@tlsomers: What do you think? If you approve, I can merge this.

tlsomers:
I haven't run across any situations where I would want to specifically wait before restarting. @Janthor have you tested this with monitor connects and disconnects to verify it does not cause multiple restarts?

Janthor:
> I haven't run across any situations where I would want to specifically wait before restarting. @Janthor have you tested this with monitor connects and disconnects to verify it does not cause multiple restarts?

I have, at least on my end it takes a lot longer for workspacer to restart than for windows to finish changing resolutions, so I haven't ran into it restarting more than once. Workspacer is installed on an SSD alongside windows so I feel it's got the best chance possible to do that.

Tests on different hardware than mine could be good though.

tlsomers:
I sadly don't have any other hardware to test on, but believe it should be alright in that case. @josteink it's approved from my perspective.

josteink (approved):

PR #361, opened 2022-09-21
Add Action-based versions of configuration extensions

tlsomers:
A small pull request to add support for configuring all current plugins with action-based extension methods.

### DIFF
--- a/src/workspacer.Bar/ConfigContextExtensions.cs
+++ b/src/workspacer.Bar/ConfigContextExtensions.cs
@@ -1,4 +1,6 @@
-﻿namespace workspacer.Bar
+﻿using System;
+
+namespace workspacer.Bar
 {
     public static class ConfigContextExtensions
     {
@@ -9,6 +11,15 @@
             context.AddLayoutProxy((layout) => new MenuBarLayoutEngine(layout, config.BarTitle, config.BarHeight));
 
             context.Plugins.RegisterPlugin(new BarPlugin(config));
+        }
+        
+        public static void AddActionMenu(
+            this IConfigContext context,
+            Action<BarPluginConfig> configAction)
+        {
+            var config = new BarPluginConfig();
+            configAction.Invoke(config);
+            AddBar(context, config);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/workspacer.Bar/ConfigContextExtensions.cs
```diff
@@ -10,5 +12,14 @@ public static void AddBar(this IConfigContext context, BarPluginConfig config =
 
             context.Plugins.RegisterPlugin(new BarPlugin(config));
         }
+        
+        public static void AddActionMenu(
```
josteink, line 16: Any reason for this method to return `void`, instead of the plugin created?

Also, it says `AddActionMenu()`, but initiates a `BarPlugin`-indtance. Should we fix naming somewhere to make this more consistent?

src/workspacer.FocusIndicator/ConfigContextExtensions.cs
```diff
@@ -1,10 +1,19 @@
...
+namespace workspacer.FocusIndicator
 {
     public static class ConfigContextExtensions
     {
         public static void AddFocusIndicator(this IConfigContext context, FocusIndicatorPluginConfig config = null)
         {
             context.Plugins.RegisterPlugin(new FocusIndicatorPlugin(config ?? new FocusIndicatorPluginConfig()));
         }
+        
+        public static void AddFocusIndicator(this IConfigContext context, Action<FocusIndicatorPluginConfig> onConfig)
```
josteink, line 12: Any reason for this method to return `void`, instead of the plugin created?

josteink:
Most of the changes look simple enough, but with some consistency errors.

If we're adding these methods, perhaps we could add that those as examples in the documentation/readme too?

src/workspacer.FocusIndicator/ConfigContextExtensions.cs
tlsomers: I maintained the same result type as the already existing extension method above. I will instead update both method return types.

josteink (approved):

josteink:
I guess at this point I'm happy. If you want to add some documentation too, that's OK by me and I'll hold merging/reviewing.

If you don't want to do that, let me know and we can get this merged.

tlsomers:
I was working on some documentation changes :)

tlsomers:
I added some text to the plugin part of the configuration page. The readme doesn't use either of the two methods.

josteink (approved):
LGTM

PR #362, opened 2022-09-22
Layout aliases - pretty names for ActiveLayoutWidget

Janthor:
The idea is creating a custom editable alias for layouts that can be shown in the ActiveLayoutWidget if the user wants to. 
With things like Nerd Fonts you can achieve layout icons or even just simple custom layout names that fit more onto your visual design.
Widget config example:
```csx
 LeftWidgets = () => new IBarWidget[]
        {
            new ActiveLayoutWidget()
            {
                UseAlias = true
            }, 
        }
```
Layout creation example:
```csx
 Func<ILayoutEngine[]> defaultLayouts = () => new ILayoutEngine[]
    {
        new TallLayoutEngine(){
            Alias = "Tall  "
        },
    };
```
Result:
image


tlsomers:
Seeing as the alias cannot be used to identify a layout, I propose renaming the field to 'DisplayedName'. Additionally, I would suggest removing UseAlias entirely in favor of just setting it to the `Name` value by default.

Janthor:
> Seeing as the alias cannot be used to identify a layout, I propose renaming the field to 'DisplayedName'. Additionally, I would suggest removing UseAlias entirely in favor of just setting it to the `Name` value by default. 

Thanks for the tips, I'll implement those tonight and update the PR.

Janthor:
I failed to implement those changes since i'm not that experienced with C# and the whole class/accessor concept (came from a C background). After countless attempts I always ended up with the DisplayedName variable being a static copy of the Name variable, so setting variables in the config file didn't change the value of DisplayedName.

I'd either like some tips on how to approach this change, or maybe a different implementation from someone who's also interested in this feature.

Janthor:
Closing this since #367 does the same but in a way cleaner way.

PR #364, opened 2022-09-26
Focus border improvements for full screen and dragging, as well as better fullscreen detection and support.

tlsomers:
- Disabled permanent border on moving a window for performance reasons.
- Force layout on movement of window which helps with apps that move into last location (ex firefox and notepad++)
- Disable layout on workspaces with full-screen window.

### DIFF
--- a/src/workspacer.FocusBorder/IFormProxy.cs
+++ b/src/workspacer.FocusBorder/IFormProxy.cs
@@ -2,17 +2,17 @@
 
 namespace workspacer.FocusBorder
 {
-    public interface IFormProxy<Form>
+    public interface IFormProxy<TForm>
     {
         /// <summary>
         /// Execute the given action in the thread that owns the form.
         /// </summary>
         /// <param name="action">The action to execute.</param>
-        public void Execute(Action<Form> action);
+        void Execute(Action<TForm> action);
 
         /// <summary>
         /// Get the underlying form. This property should only be used to read values.
         /// </summary>
-        public Form Read { get; }
+        TForm Read { get; }
     }
 }
--- a/src/workspacer.FocusBorder/IFormProxy.cs
+++ b/src/workspacer.FocusBorder/IFormProxy.cs
@@ -1,0 +1,18 @@
+﻿using System;
+
+namespace workspacer.FocusBorder
+{
+    public interface IFormProxy<Form>
+    {
+        /// <summary>
+        /// Execute the given action in the thread that owns the form.
+        /// </summary>
+        /// <param name="action">The action to execute.</param>
+        public void Execute(Action<Form> action);
+
+        /// <summary>
+        /// Get the underlying form. This property should only be used to read values.
+        /// </summary>
+        public Form Read { get; }
+    }
+}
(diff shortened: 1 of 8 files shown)


src/workspacer.FocusBorder/IFormProxy.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace workspacer.FocusBorder
+{
+    public interface IFormProxy<Form>
+    {
+        /// <summary>
+        /// Execute the given action in the thread that owns the form.
+        /// </summary>
+        /// <param name="action">The action to execute.</param>
+        public void Execute(Action<Form> action);
```
josteink, line 11: `public` modifier is not needed for public interface.

src/workspacer.FocusBorder/IFormProxy.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        /// <summary>
+        /// Get the underlying form. This property should only be used to read values.
+        /// </summary>
+        public Form Read { get; }
```
josteink, line 16: `public` modifier is not needed for public interface.

src/workspacer.FocusBorder/IFormProxy.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace workspacer.FocusBorder
+{
+    public interface IFormProxy<Form>
```
josteink, line 5: In C# when having type-arguments, it's customary to prefix them with `T` to avoid confusion inside the interface or class which depends on them.

`TForm` would be better here. Would it also make sense to apply any constraints to what can be a form in this context?

src/workspacer.Native/Native/WindowsWindow.cs
```diff
@@ -151,6 +151,10 @@ public bool CanLayout
         public bool IsFocused => Win32.GetForegroundWindow() == _handle;
         public bool IsMinimized => Win32.IsIconic(_handle);
         public bool IsMaximized => Win32.IsZoomed(_handle);
+        public bool IsFullscreen =>
+            (~Win32.GetWindowStyleLongPtr(_handle))
+                .HasFlag(Win32.WS.WS_OVERLAPPED | Win32.WS.WS_BORDER | Win32.WS.WS_THICKFRAME);
```
josteink, line 156: I'm not against inlined methods when it makes sense, but I think this is getting dense to the point that it's hard to read what it actually does.

I'm also curious about the use of the unary bitwise complement operator here.

Wouldn't it be more C# convention to extract the value, and then instead check for the absence of those flags?

src/workspacer/Workspace/WorkspaceManager.cs
```diff
@@ -417,7 +417,8 @@ public void UpdateWindow(IWindow window, WindowUpdateType type)
                         }
                     }
 
-                    if (type == WindowUpdateType.Move)
+                    // Only change the current window if we are actively moving it.
+                    if (type == WindowUpdateType.Move && window.IsMouseMoving)
```
josteink, line 421: This seems like a unrelated change... Was this actually required, or is it just something else you discovered and decided to fix at the same? :smile:

josteink:
Some small nitpicks, but otherwise it looks good to me :smile:

src/workspacer.Native/Native/WindowsWindow.cs
tlsomers: The main reason for using the bitwise operators was for performance, using a single invert and method call is significantly faster than 3 separate method calls. Also for setting and clearing flags in C# you have to use bitwise operators anyways.

josteink (approved):
LGTM

PR #367, opened 2022-09-30
Allow overriding the name of layouts. Similar to the alias PR.

tlsomers:
@Janthor This is how I would do it, just allow setting the name field

### DIFF
--- a/src/workspacer.Bar/Layouts/MenuBarLayoutEngine.cs
+++ b/src/workspacer.Bar/Layouts/MenuBarLayoutEngine.cs
@@ -8,7 +8,12 @@ namespace workspacer.Bar
         private string _title;
         private int _offset;
         private ILayoutEngine _inner;
-        public string Name => _inner.Name;
+
+        public string Name
+        {
+            get => _inner.Name;
+            set => _inner.Name = value;
+        }
 
         public MenuBarLayoutEngine(ILayoutEngine inner, string title, int offset)
         {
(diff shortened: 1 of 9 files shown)


Janthor:
I thought there was a reason for the name variable being static(referencing it from other places maybe), if it wouldn't hurt to change it it'd probably be the best solution.

tlsomers:
It's only used in LayoutName which is a property so changes automatically, and LayoutName is only used for the bar item, so it shouldn't cause any harm

Janthor (commented):
I dunno how much my review is worth, but I tried it locally and it's failing to run when I try to set the name from the config.
```
Microsoft.CodeAnalysis.Scripting.CompilationErrorException: '(143,13): error CS0200: Property or indexer "DwindleLayoutEngine.Name" cannot be assigned to -- it is read only'
```
I honestly think something is wrong with visual studio and it's not properly compiling the changes for some reason, but I also can't get it to compile correctly even after cleaning and rebuilding the solution multiple times

This is how I was declaring the name in the config:
```csx
    Func<ILayoutEngine[]> defaultLayouts = () => new ILayoutEngine[]
    {
        new DwindleLayoutEngine(1,0.7,0.03){
            Name = "侀"
        },
    }
```

tlsomers:
@Janthor remember to rebuild the entire solution, and make sure that your r directives point to the build output. It's just an outdated dll

Janthor:
> @Janthor remember to rebuild the entire solution, and make sure that your r directives point to the build output. It's just an outdated dll

Oooooh forgot to change out the DLLs in the config, sorry lmao

Janthor (approved):
Confirmed working 
image

josteink (approved):
LGTM

PR #370, opened 2022-10-03
Bar margins - Ability to vertically position bar text

Janthor:
Pretty self-explanatory. Currently all text rendered to the bar is snapped to the top, that creates uneven empty space when the font size doesn't match the current bar size. To circumvent that, a new parameter to set the bar margin is created. 
I tried messing with the bar rendering itself to make it automatically center but couldn't find a neat solution, so I settled on making it a custom parameter.

how to configure:
```csx
    context.AddBar(new BarPluginConfig()
    {
        //* Bar config
        ...
        BarMargin = 5,
        ...
    }
```
before:
image

after:
image


Janthor:
I need to figure out how to make github not show a bunch of unnecessary commits in my PRs 😒

josteink (approved):
LGTM

josteink:
Or more simply, just always create your PRs based on latest commits in workspacer git master :smile:

PR #371, opened 2022-10-06
Add behavior to spawn windows in the master slot (top of the list)

Janthor:
Pretty self explanatory. For my needs, I almost always want to focus on the new windows I open, so I added the behavior to the configuration context for spawning windows as the first items in the linked list (master slot), instead of the current behavior that always spawns them on the bottom of the list. 

I am very bad at naming variables, so I accept suggestions on a better name for this config variable.

Spawning Explorer with ``context.NewWindowAsMaster <- false`` (old behavior)
image
Spawning Explorer with ``context.NewWindowAsMaster <- true`` (new behavior option)
image

### DIFF
--- a/src/workspacer.Shared/IConfigContext.cs
+++ b/src/workspacer.Shared/IConfigContext.cs
@@ -23,8 +23,12 @@ namespace workspacer
 
         IMonitorContainer MonitorContainer { get; set; }
         bool CanMinimizeWindows { get; set; }
-        bool NewWindowAsMaster { get; set; }
-        
+        WindowOrder NewWindowOrder { get; set; }
+        public enum WindowOrder
+        {
+            NewWindowsLast, // default
+            NewWindowsFirst
+        }
         string ConfigDirectory { get; }
 
         /// <summary>
--- a/src/workspacer.Shared/IConfigContext.cs
+++ b/src/workspacer.Shared/IConfigContext.cs
@@ -24,11 +24,7 @@ namespace workspacer
         IMonitorContainer MonitorContainer { get; set; }
         bool CanMinimizeWindows { get; set; }
         WindowOrder NewWindowOrder { get; set; }
-        public enum WindowOrder
-        {
-            NewWindowsLast, // default
-            NewWindowsFirst
-        }
+
         string ConfigDirectory { get; }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/workspacer.Shared/Workspace/Workspace.cs
```diff
@@ -73,7 +73,15 @@ public void AddWindow(IWindow window, bool layout = true)
...
 
-                _windows.Add(window);
+                if (_context.NewWindowAsMaster)
+                {
+                    _windows.Insert(0, window);
+                }
+                else
+                {
+                    _windows.Add(window);
+                }
```
josteink, line 83: This change looks simple enough, and from what I can tell changes the order of which windows gets added.

How about, instead of a boolk, making this an enum? Then one can make both the property-name and property-value more descriptive.

A quick suggestions here, without too much thought into it (yet):

````c#
WindowOrder WindowAddingOrder {get; set; }

public enum WindowOrder
{
    NewWindowsLast, // default
    NewWindowsFirst
}
````

Wouldn't that be a more clear description of what this change involves?

Comment (rickbutton) [reply src/workspacer.Shared/Workspace/Workspace.cs:83]: maybe `WindowOrder => NewWindowOrder` but +1

**Comment by Janthor** [comment]
@josteink @rickbutton done

File: src/workspacer.Shared/IConfigContext.cs [hunk 2]
```diff
@@ -23,7 +23,12 @@ public interface IConfigContext
 
         IMonitorContainer MonitorContainer { get; set; }
         bool CanMinimizeWindows { get; set; }
-        
+        WindowOrder NewWindowOrder { get; set; }
+        public enum WindowOrder
+        {
+            NewWindowsLast, // default
+            NewWindowsFirst
+        }
```
josteink, line 31: It would be better if this enum was a standalone entity and not nested within another.

josteink (changes requested):
Almost there. One last thing 🙂

src/workspacer.Shared/IConfigContext.cs
Janthor: did it 👍

josteink (approved):

josteink:
Thanks for the PR. Consider it merged!

Issue #368, opened 2022-10-01
Discord window isn't getting managed


Janthor:
I can't pinpoint which of the latest PR's caused it, but Discord isn't getting moved/resized by workspacer anymore. It does get the focus indicator though.
Here are the window's properties:
```
CLASS                    Executable      Program name
Chrome_WidgetWin_1	Discord.exe	Discord	
```
I tested reverting back to the stable version and it went back to being properly managed, I am also certain that it was working on my PR #359

Janthor:
going over the changes, I suspect it is caused by one of these two changes on the focus border improvements:
[WindowManager.cs](https://github.com/workspacer/workspacer/commit/b24af60f19612ef945155faea50ad77bf2bac402#diff-a6493c28818b8d0cb1eda7ac526dafd8915e8942e276b9139f9eae0b850fa011)
[WindowsWindow.cs](https://github.com/workspacer/workspacer/commit/b24af60f19612ef945155faea50ad77bf2bac402#diff-aef4e4e76ca52011fbfbf69044568fdf97eb2e3b41e8690e4c8d881ff50690f0)

Janthor:
After further inspection, this is the cause of the issue, introduced in #364 :
https://github.com/workspacer/workspacer/blob/b24af60f19612ef945155faea50ad77bf2bac402/src/workspacer.Native/Native/WindowsWindow.cs#L155-L163

I'm trying to find other ways of determining if a window is fullscreen, since some non-fullscreen apps like Discord can have these exact style properties

josteink:
CC: @tlsomers

tlsomers:
This fix will not work when the bar and taskbar are both off, as a single window will be seen as Fullscreen. Also I was unable to reproduce the discord issue. Could you provide a screenshot @Janthor

Janthor:
I think I tried maximized windows and it didn't trigger the Fullscreen code, if I'm not mistaken windows tries to render a drop shadow on maximized windows that makes the window rect bigger than the screen rect, and those shadows gets disabled on a true Fullscreen window. I did try it with a two monitor setup though, need to test if the same is true for a single monitor.

PR #373, opened 2022-10-07
Fix fullscreen detection

Janthor:
Current implementation has some false positives (I.E. Discord) as per issue #368 . This PR proposes new ways of detecting if an app is running in fullscreen mode. The idea for this is checking if the window's rectangle is the same size as the rectangle from the monitor it's contained in.


josteink (approved):

josteink:
LGTM

PR #391, opened 2022-11-08
updated docs with alternative config location

YesSeri:
I have added the alternative location of the config file to the quickstart docs.


docs/content/quickstart.md
```diff
@@ -38,8 +38,11 @@ Workspacer loads a default configuration with almost everything you'd need, but
 
 A configuration file can be generated by right clicking the ```workspacer``` system tray icon, and selecting the option to create an example configuration in:
 
+The file gets created in one of the two following places
```
josteink, line 41: I think having a `:` at the end here would be better, but otherwise this is good :smile:

josteink (approved):

PR #417, opened 2023-02-22
Feat/navigate across monitor

blalande:
Add two functions to workspace manager to allow for navigation across all monitors


josteink:

In general the idea seems useful, and superficially the code looks good.

I'll try to get this properly reviewed and merged as soon as I have time!

src/workspacer/Workspace/WorkspaceManager.cs
```diff
@@ -28,6 +28,73 @@ public WorkspaceManager(IConfigContext context)
...
+                    }
+                    didFocus = true;
+                    break;
+                }
+            }
+            if(!didFocus && windows.Count > 0)
+            {
+                // shouldn’t happen.
+                windows[0].Focus();
+            }
```
josteink, line 63: If this shouldnt happen, do we really need this code? :)

src/workspacer/Workspace/WorkspaceManager.cs
```diff
@@ -28,6 +28,73 @@ public WorkspaceManager(IConfigContext context)
...
```
josteink, line 96: If this shouldnt happen, do we really need this code? :)

src/workspacer/Workspace/WorkspaceManager.cs
```diff
@@ -28,6 +28,73 @@ public WorkspaceManager(IConfigContext context)
...
+                if (window.IsFocused)
+                {
+                    if (i + 1 == windows.Count)
+                    {
+                        //go to next monitor that has windows
+                        do 
+                        {
+                            this.SwitchFocusToNextMonitor();
+                        }
+                        while(!FocusedWorkspace.ManagedWindows.Any());
```
josteink, line 48: In theory, if you have no windows open and trigger this method, won't this code loop forever?

I think we may benefit by extracting this code into its own function (`SwitchFocusToNextMonitorWithWindows()` ?) where we can implement a better check without complicating this method.

src/workspacer/Workspace/WorkspaceManager.cs
```diff
@@ -28,6 +28,73 @@ public WorkspaceManager(IConfigContext context)
...
+                var window = windows[i];
+                if (window.IsFocused)
+                {
+                    if (i  == 0)
+                    {
+                        //go to previous monitor that has windows
+                        do {
+                        this.SwitchFocusToPreviousMonitor();
+                        }
+                        while(!FocusedWorkspace.ManagedWindows.Any());
```
josteink, line 81: In theory, if you have no windows open and trigger this method, won't this code loop forever?

I think we may benefit by extracting this code into its own function (`SwitchFocusToPreviousMonitorWithWindows()` ?) where we can implement a better check without complicating this method.

josteink:
Hey there and sorry for the delay.

I finally got around to reviewing this PR and I have a few small comments.

If you get those issues resolved, this looks like a nice addition and I would be happy to merge it.

src/workspacer/Workspace/WorkspaceManager.cs
blalande: yes I didn’t see that. Will update

src/workspacer/Workspace/WorkspaceManager.cs
blalande: yes I didn’t see that. Will update

src/workspacer/Workspace/WorkspaceManager.cs
blalande: I copied the code from the window class and didn’t think too much about it, but yeah it shouldn’t be needed

src/workspacer/Workspace/WorkspaceManager.cs
blalande: I copied the code from the window class and didn’t think too much about it, but yeah it shouldn’t be needed

blalande:
I’m just thinking that I should also have the same method for moving windows across monitor in the same way. I’ll update the MR with that.

blalande:
Hello,

Sorry life's been a bit full. I'll try to work on that when I can, I havent
forgotten about it :)

Aareon:
This PR is in need of help. If anyone wants to own this, please let us know.

Issue #431, opened 2023-04-12
feature request: static configuration


jshield:
For the sake of performance and ease of debugging it would be nice if workspacer could look for a DLL named workspacer.Config.dll under %USERPROFILE%/.workspacer and if present load the assembly and use that for configuration, then falling back to compiling and executing a CSX script.

could look for a class named WorkspacerConfiguration that exports a method that returns the expect builder.

jshield:
hey mate thanks for the reply, by performance I mean the time it takes for the cscript to compile is non trivial particularly when I have a number of custom widgets and for whatever reason need to restart workspacer. perhaps if the resulting compiled cscript assembly could be cached somewhere it might not be an issue.

as for what happens, I'd imagine it would be the same as what happens now with the cscript if there is an API break it'll break, however it is a lot easier for me to automate testing my configuration still compiles using standard .net tools and github actions than relying on a custom cscript evaluator and having to launch workspacer on an agent in an interactive session and then script commands to terminate it.

a potential mitigation might be to release the configuration surface as a set of abstractions to nuget similar to how dotnet extensions like IConfiguration, DependencyInjection are released. those shouldn't change often, and can be versioned according to semver, although that would be additional maintenance burden and probably would only need to be done as a last resort.

I ended up rolling something that works for me so far, I've attached a PR for reference. it probably needs further work if it were to be merged into unstable.

josteink:
> as for what happens, I'd imagine it would be the same as what happens now with the cscript if there is an API break it'll break

The problem is the csharp-script gets compiled against whatever running assembly is running (so it's guaranteed to be the right one). while once compiled to a static DLL, the .NET runtime will link the compiled DLL to a specific versions of referenced DLLs.

This could create issues with loading the static configuration DLL after new versions are published and installed.

It's not impossible to handle. For instance, one can override the [AssemblyLoadContext](https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext) when loading the configuratio, but like I said it's going to involve at least a little bit of work.

> a potential mitigation might be to release the configuration surface as a set of abstractions to nuget similar to how dotnet extensions like IConfiguration, DependencyInjection are released. those shouldn't change often, and can be versioned according to semver, although that would be additional maintenance burden and probably would only need to be done as a last resort.

Yeah. That sounds like a bigger kind of job, for IMO a very niche need. I'd rather not go down that road :)

[...]

PR #432, opened 2023-04-12
feat: load configuration from assembly if present

jshield:
implements a version of #431

### DIFF
--- a/src/workspacer/Config/ConfigHelper.cs
+++ b/src/workspacer/Config/ConfigHelper.cs
@@ -10,6 +10,8 @@ namespace workspacer
     public static class ConfigHelper
     {
         private static readonly string ConfigFileName = "workspacer.config.csx";
+        private static readonly string ConfigAssemblyName = "workspacer.Configuration.dll";
+        private static readonly string ConfigLoadedAssemblyName = "workspacer.Configuration.dll.loaded";
 
         private static string GetPathInUserFolder(string file)
         {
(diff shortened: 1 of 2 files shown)


src/workspacer/Config/ConfigHelper.cs
```diff
@@ -58,14 +60,52 @@ private static string LoadConfig()
...
+                {
+                    // prevent locking issues when copying the configuration assembly while workspacer is running.
+                    var copiedAssembly = GetPathInUserFolder(ConfigLoadedAssemblyName);
+                    if (File.Exists(copiedAssembly))
+                    {
+                        File.Delete(copiedAssembly);
+                    }
+                    File.Copy(configurationAssemblyPath, copiedAssembly);
+                    configurationAssemblyPath = copiedAssembly;
+                    _configurationAssembly = Assembly.LoadFile(configurationAssemblyPath);
```
josteink, line 83: From my understanding of things, `Assembly.LoadFile` actually caches the loaded assembly.

This would mean that compiling a new config and doing a workspacer reload wont actually load the new configuration.

Isn't that a bit counter-intuitive?

jshield: ah yeah good point I should probably create a new AssemblyLoadContext, and dispose/create anew.
I don't tend to use reload tbh as I am usually running it with a debugger attached, and actively making code changes.

I was under the impression that while it is indeed cached it keeps a lock on the assembly file on disk, I noticed that it would remain locked when I was running in release mode, perhaps I should read the assembly into a memory stream and load it from there instead of doing the copy to disk

josteink: > I don't tend to use reload tbh as I am usually running it with a debugger attached, and actively making code changes

In general, I think when making a config it's common to try different things interactively until you find what works. It would be nice to be able to support that flow.

> perhaps I should read the assembly into a memory stream and load it from there instead of doing the copy to disk

That's probably a simpler approach, yeah.

PR #439, opened 2023-04-26
Using UserEnviromentVariables to set the folder Path

dantearaujo1:
Now we can set a UserEnviromentVariable to move .config/workspacer folder.


josteink:
Change looks pretty harmless to me, and I'm OK with the implementation as is.

But you may want to document it? Try to find somewhere in [the docs](https://github.com/workspacer/workspacer/tree/master/docs) and add it there too? :)

dantearaujo1:
I'll do my best

josteink (approved):
LGTM

josteink:
Thanks for the patch. Consider it merged!

PR #480, opened 2023-09-04
Added Center Widgets and Nord Theme Colors

Entharia:
Hello, I thought it would be nice to have center widgets and some more color choices.
Also, I added a property to change the color of a workspace.

### DIFF
--- a/src/workspacer.Bar/BarForm.cs
+++ b/src/workspacer.Bar/BarForm.cs
@@ -10,11 +10,13 @@ namespace workspacer.Bar
         private IMonitor _monitor;
         private BarPluginConfig _config;
         private System.Timers.Timer _timer;
-        private FlowLayoutPanel leftPanel;
-        private FlowLayoutPanel rightPanel;
+        private FlowLayoutPanel _leftPanel;
+        private FlowLayoutPanel _rightPanel;
+        private FlowLayoutPanel _centerPanel;
 
         private BarSection _left;
         private BarSection _right;
+        private BarSection _center;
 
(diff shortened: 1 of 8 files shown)


josteink:

It's a very big PR, so clearly you've spent a lot of time and effort on it. That's appreciated.

Now for some bad news 😄 

* This PR adds a center bar, but also changes the API contract slightly. This may be a breaking change for some people, so if that's something we're going to do, we should find a way to do it in a backwards compatible wau.
* You've added a new theme. [This Nord theme?](https://www.nordtheme.com) It might be great, but it's also now the default, not available as an option or configuration. Is that intentional?
* Improved the PomodorWidget. Whee!
* Done some other minor code reformatting and refactoring. It may make the code more readable. It may have introduced bugs. Proper code-review will be required!

Did I miss something? 

Usually on Github when you submit a PR, you try to make it as a minimal as possible, and prefer to submit several parallell/successive PRs where each change you want to implement can be reviewed in isolation.

Review requires the reviewer to understand the code, so less code at a time is better for a proper review. And a proper review often suggests code-changes. In which case, having a small PR means less code to change 😉 

I wouldn't mind some of these changes, but to review them effectively, I think you should submit them as separate PRs and then we can tackle them one by one.

I see you've put effort into this, so I hope doing it like this instead is OK for you.

Entharia:
Hi, the pomodoro stuff was supposed to be its own PR.
I will try to make my changes without changing the API contract.
I will close this PR until im done with the new changes

